PR #6, opened 2020-04-16
Added document planning metaflow migration

haakondr:
I attempted to make a document identifying the mapping between concepts in metaflow/tsb and ODM2 (and Roalds adapted version with trackresults)

PS: this is incomplete, but thought I'd just make a PR to start discussion.

### DIFF
--- a/docs/metaflow-migration.md
+++ b/docs/metaflow-migration.md
@@ -1,0 +1,17 @@
+# Metaflow/tsb migration
+
+The purpose of this document is to describe the relationship between the, at time of writing, the current metadata structure in metaflow and ODM2
+
+## mapping
+
+Below is a mapping between concepts in metaflow/tsb and their corresponding table in ODM2
+
+
+|metaflow ttype | tsb table | ODM2 metadata     | ODM2 data             |
+|---|---|---|---|
+| vessel      |      |   |   |
+| component      |      |   |   |
+| gpstrack      | track     | samplingfeatures  | trackresultlocations  |
+| tseries       | ts        | trackresults      | trackresultvalues     |
+| qctseries     | flag      |               |                   |
+


docs/metaflow-migration.md
```diff
@@ -0,0 +1,17 @@
...
+Below is a mapping between concepts in metaflow/tsb and their corresponding table in ODM2
+
+
+|metaflow ttype | tsb table | ODM2 metadata     | ODM2 data             |
+|---|---|---|---|
+| vessel      |      |   |   |
+| component      |      |   |   |
+| gpstrack      | track     | samplingfeatures  | trackresultlocations  |
+| tseries       | ts        | trackresults      | trackresultvalues     |
+| qctseries     | flag      |               |                   |
```
haakondr, line 16: I've understood it such that there is no concept for qc timeseries in ODM2?

docs/metaflow-migration.md
```diff
@@ -0,0 +1,17 @@
...
+## mapping
+
+Below is a mapping between concepts in metaflow/tsb and their corresponding table in ODM2
+
+
+|metaflow ttype | tsb table | ODM2 metadata     | ODM2 data             |
+|---|---|---|---|
+| vessel      |      |   |   |
+| component      |      |   |   |
+| gpstrack      | track     | samplingfeatures  | trackresultlocations  |
```
haakondr, line 14: Do we have any concept that maps to our current component and vessel hierarchy? Does the action (Instrument deployment) of setting up a ferrybox in a ship sort of map to our old concept of vessel?

haakondr:
I also have some additional questions:

- are paths stored? If so, where?
- results.valuecount is this supposed to be incremented on insert?
- the code at https://github.com/NIVANorge/nivacloud/commit/2e8dd272e9a2e4b0337ac728fc6c7392a07d0bf8#diff-dd7538b086903d6a5bceabbf01e7bac0R153 seems to insert all data for all resultIds in the dict. Is this correct?

haakondr:
could we use PointCoverageResultValue instead of our custom trackresult concept?

I still think that storing location per measurement/resultvalue is the way to go in order to get proper query performance. At least we should weight the pros/cons for either approach. 

Question is, how many extra gigabytes this would add to our volume, and whether or not we can achieve geo based queries with the trackresult approach.

http://odm2.github.io/ODM2/schemas/ODM2_2_0_0/tables/ODM2Results_PointCoverageResultValues.html

haakondr:
I am trying to understand how I can map from path to result. Have I understood this correctly?

Some person performs an action (Instrument deployment) to install a ferrybox. This results in multiple sampling features (CTD, Ships track, Water quality station?). Each sampling features have 1 result.

If the above is correct, could we store the timeseries path as samplingfeatures.samplingfeaturecode ?

We would have to solve the backwards compatibility of old paths somehow, but lets not worry too much about that for now

Issue #9, opened 2020-04-20
Associate actions with logged in port-user


haakondr:
All actions being made to the ODM2 API should be associated with a logged in user, instead of a hardcoded one. We need to pick up the niva-port user and map this to a person entity in the people table.

Currently, we store users in niva-port with the following identifiers:

- id (primary key, sequential number)
- uid (external user id from azure AD, google oauth, etc)

Niva-port passes along niva-port user.id (primary key) as http header "User-Id" on proxied requests. All port apps will receive this http  header as part of all requests.

## Tasks

- pick up logged in user and associate actions with user in people table 
  - if user does not exist in ODM2 database, create it
- emulate niva-port behavior in dev mode

see https://github.com/ODM2/ODM2/blob/master/doc/ODM2Docs/core_people.md for explanation of the ODM2 "people".
see related issue in niva-port: https://github.com/NIVANorge/niva-port/issues/188

PR #11, opened 2020-04-22
Added function to extract nivaport user from headers + api tests #9

haakondr:
Creates a odm2 person/affiliation if it does not exist. This function is not yet used, but is intended to be called with Niva-User header from requests. Niva-User header is added to requests
when proxied behind niva-port

##  function create_or_get_user:

Takes base64-encoded json string and parses it into a user object. Then
tries to find user based on user.id. If user does not exist, it inserts
it into the database based on user object object header

## misc changes:

- added tests which boots up a postgres instance dedicated for testing.
Wipes all data before each test (custom truncate all function)
  - had some trouble wiping the database with the odm2 user. Using
postgres user instead as a workaround (probably ok for tests?):

```
asyncpg.exceptions.InsufficientPrivilegeError: must be owner of
hypertable "trackresultlocations"
```

TODO: a more proper setup of intial metadata in tests. Just doing 3
ad-hoc inserts now. Need some basic controlled vocabularies++. Just did
this very quickly now
TODO: properly populate affiliations++. Scoped this out for now

@jemmima: added you in case you'd want to do some reviewing
@RmStorm: I did not quite get the tests to run on circleci, but they do pass locally. Don't get time to resolve this now, but hopefully this is enough to do a code review


src/odm2_postgres_api/queries/user.py
```diff
@@ -0,0 +1,78 @@
...
+    middle_name = " ".join(mid) if mid else None
+    name_tuple = (head, middle_name, tail)
+    if not head or not tail:
+        raise Exception("could not parse user fullname")
+    return name_tuple
+
+
+def get_nivaport_user(user_id_header: str) -> NivaPortUser:
+    user_json = json.loads(b64decode(user_id_header))
+    return NivaPortUser(**user_json)
```
haakondr, line 33: taking the header values from niva-port and convert to a object.

This is a very simple approach with base64 encoding. May consider opting for some jwt tokens or something for internal networking, but thought I'd do it simple for now

src/odm2_postgres_api/queries/user.py
```diff
@@ -0,0 +1,78 @@
...
+
+
+
+class StoredPerson(BaseModel):
+    person: People
+    affiliation: Affiliations
+    external_identifier: PersonExternalIdentifiers
```
haakondr, line 40: don't really need all this information, but just exposing everything we insert for testing purposes. May look into a more practical wrapper object if needed

src/odm2_postgres_api/queries/user.py
```diff
@@ -0,0 +1,78 @@
...
+    if stored_user:
+        return stored_user
+    else:
+        async with conn.transaction():
+            name = full_name_to_split_tuple(user.displayName)
+            people_row = await conn.fetchrow(
+                "INSERT INTO odm2.people (personfirstname, personmiddlename, personlastname) "
+                "VALUES ($1, $2, $3) "
+                "RETURNING *", *name)
+            person_created = People(**people_row)
```
haakondr, line 59: can we somehow set default schema on the db connection/pool?

tests/integration_test_fixtures.py
```diff
@@ -0,0 +1,49 @@
...
+    db_port = os.environ["TIMESCALE_ODM2_SERVICE_PORT"]
+    user = os.environ["POSTGRES_USER"]
+    password = os.environ["POSTGRES_PASSWORD"]
+    db_name = os.environ["ODM2_DB"]
+    return await asyncpg.create_pool(user=user, password=password,
+                                     host=db_host, port=db_port, database=db_name)
+
+
+@pytest.fixture
+async def clear_db():
```
haakondr, line 36: should this be scope="function" in order to run before each test? have not tested this, as I had only one test

tests/integration_test_fixtures.py
```diff
@@ -0,0 +1,49 @@
...
+        # TODO: move into general metadata (this should be the same in all environments..)
+        await connection.fetchrow("INSERT INTO odm2.cv_organizationtype "
+                                  "(term, name) "
+                                  "VALUES ('institute', 'Institute');")
+        await connection.fetchrow("INSERT INTO odm2.organizations "
+                                  "(organizationtypecv, organizationcode, organizationname) "
+                                  "VALUES ('Institute', 'niva', 'Norwegian institute for Water Research');")
+        await connection.fetchrow("INSERT INTO odm2.externalidentifiersystems "
+                                  "(identifiersystemorganizationid, externalidentifiersystemname) "
+                                  "VALUES ((select organizationid from odm2.organizations where organizationcode='niva'), 'niva-port');")
```
haakondr, line 49: quick and dirty, should probably look into a more solid solution

tests/test_utils.py
```diff
@@ -0,0 +1,28 @@
...
+            logging.info(f"Successfully truncated table {table}")
+            # except Exception as e:
+            #     failed.insert(table)
+
+        if len(failed) > 0:
+            await _truncate_tables(tables=failed, iterations=iterations - 1)
+
+    rows = await connection.fetch("select * from pg_tables where schemaname=$1;", schema)
+    tables = [t["tablename"] for t in rows]
+    await _truncate_tables(schema=schema, tables=tables)
```
haakondr, line 28: This function clears our db data without dropping schema information. Not sure how this behaves with foreign key constraints, that's why I added the iteration part, but commented it out. 

Surely there must be libraries that does this, but I did not spend any time googling for it. Intended to be ran before each test

src/odm2_postgres_api/queries/user.py
```diff
@@ -0,0 +1,80 @@
...
+    User object passed from niva-port as a base64 encoded json string
+    """
+    id: str
+    email: str
+    roles: List[str]
+    displayName: str
+
+
+def full_name_to_split_tuple(fullname: str) -> Tuple[str, Optional[str], str]:
+    head, *mid, tail = fullname.strip().split(" ")
```
RmStorm, line 23: if there is only 1 name in fullname this will throw a valueerror and the error handling on line 26 will never be hit:
alternative1:
```python
    try:
        head, *mid, tail = fullname.strip().split(" ")
        middle_name = " ".join(mid) if mid else None
        name_tuple = (head, middle_name, tail)
        return name_tuple
    except ValueError:  # destructuring can lead to ValuErrors if there are not enough names in the string
        raise Exception("could not parse user fullname")
```
alt2
```python
    names = fullname.strip().split(" ")
    if len(names) == 1:
        raise Exception(f"Need at least first and last name: {names=}")
    if len(names) == 2:
        return names[0], None, names[-1]
    else:
        return names[0], " ".join(names[1:-1]), names[-1]
```

RmStorm: also, hitting that error conditions has to be part of one of the tests

src/odm2_postgres_api/queries/user.py
```diff
@@ -0,0 +1,80 @@
...
+
+class StoredPerson(BaseModel):
+    person: People
+    affiliation: Affiliations
+    external_identifier: PersonExternalIdentifiers
+
+
+async def create_or_get_user(conn: asyncpg.connection, user_id_header: str) -> StoredPerson:
+    user = get_nivaport_user(user_id_header=user_id_header)
+    stored_user = await conn.fetchrow("SELECT p.* from odm2.people p inner join odm2.personexternalidentifiers pei "
```
RmStorm, line 44: I don't see how affiliation is inserted here

src/odm2_postgres_api/queries/user.py
RmStorm: yes, it's only the postgres user that doesn't have that done. The read only user and the odm2 user both have that.

tests/integration_test_fixtures.py
RmStorm: It should default to scope=function

src/odm2_postgres_api/queries/user.py
haakondr: Good catch.

I think since people.personfirstname and people.personlastname are both required fields in odm2, I'd rather throw an error if we'd get just one name. Most likely this is an error somewhere else in the system, and I'd prefer it to blow up.

I added a check for this and throw ValueError (what is typical exception to throw when a function receives bad input?)

PS: I am not printing the name as part of exception, I think we should avoid that due to GDPR restrictions. I think the "right to be deleted" can be a hassle if we end up having names lying around in our logs. That would probably be taken care of by log retention policies, but still..

src/odm2_postgres_api/queries/user.py
haakondr: https://github.com/NIVANorge/odm2-postgres-api/pull/11/files/bb02898d6c43070f8f8acf6f734348ceebbc8c45#diff-593ba63d49b31c17864b5a1b9f6e3b0dR73-R77

haakondr:
hmm, failing due to path issues:

```
>           with open(Path(__file__).parent / 'ODM2_for_PostgreSQL.sql') as my_file:
 
E           FileNotFoundError: [Errno 2] No such file or directory: '/home/circleci/project/venv/lib/python3.7/site-packages/odm2_postgres_api/utils/ODM2_for_PostgreSQL.sql'
 ```

I guess this happens when we do "pip install ." ? @RmStorm do you have any suggestions as to how we should load this file properly?


https://app.circleci.com/pipelines/github/NIVANorge/odm2-postgres-api/57/workflows/a9c4e3ac-686c-4bcf-8a23-12675b8050a2/jobs/158

**Comment by haakondr** [comment]
> hmm, failing due to path issues:
> 
> ```
> >           with open(Path(__file__).parent / 'ODM2_for_PostgreSQL.sql') as my_file:
>  
> E           FileNotFoundError: [Errno 2] No such file or directory: '/home/circleci/project/venv/lib/python3.7/site-packages/odm2_postgres_api/utils/ODM2_for_PostgreSQL.sql'
> ```
> 
> I guess this happens when we do "pip install ." ? @RmStorm do you have any suggestions as to how we should load this file properly?
> 
> https://app.circleci.com/pipelines/github/NIVANorge/odm2-postgres-api/57/workflows/a9c4e3ac-686c-4bcf-8a23-12675b8050a2/jobs/158

Think I solved this by adding this to setup.py:

```
    package_data={'': ['**/ODM2_for_PostgreSQL.sql']},
```

PR #14, opened 2020-04-27
deploy to production

RmStorm:


src/odm2_postgres_api/utils/populate_with_general_metadata.py
```diff
@@ -0,0 +1,304 @@
...
+
+import requests
+
+from nivacloud_logging.log_utils import setup_logging
+
+
+people_and_organizations = {"people": [{
+    "personfirstname": "Roald",
+    "personlastname": "Storm"
+}, {
```
haakondr, line 11: we should delete user/affiliation

tests/integration-test-docker-compose.yml
```diff
@@ -0,0 +1,12 @@
+version: '3.5'
+
+services:
+  timescale_odm2:
+    # this version is the current timescaledb version running in production at the time of writing
+    # Remember to change the image version in both tsb/docker-compose.yml and the nivacloud timescale deployment
+    image: timescale/timescaledb-postgis:1.7.0-pg12
```
haakondr, line 7: I see that I setup a newer version of timescale/postgres for testing.. we will hopefully setup this in prod soon so I suggest we leave it for now

PR #18, opened 2020-04-30
Feature/begroing odm2 inserts

RmStorm:


src/odm2_postgres_api/schemas/schemas.py
```diff
@@ -1,13 +1,48 @@
...
+            if values[offset_name] != date_time.utcoffset().seconds / 3600:  # type: ignore
+                raise ValueError(f"conflicting utcoffset on '{datetime_name}.utcoffset().seconds/3600="  # type: ignore
+                                 f"{date_time.utcoffset().seconds / 3600}' and "
+                                 f"'values[{offset_name}]={values[offset_name]}'")
+        else:
+            if values[offset_name] != 0:
+                raise ValueError(f"'{offset_name}' should be 0 when no timezone on '{datetime_name}' is supplied")
+        return date_time.replace(tzinfo=None)
+
+    return check_utc_offset
```
haakondr, line 23: I am not quite following this. If we just require a datetime object and create a datetime validator that checks if tzinfo is not None?

as long as tzinfo is provided we have everything we need?

src/odm2_postgres_api/schemas/schemas.py
```diff
@@ -384,13 +441,17 @@ class ResultSharedBase(BaseModel):
     zlocation: Optional[float]
     zlocationunitsid: Optional[int]
     spatialreferenceid: Optional[int]
+    valuedatetimeutcoffset: int
+    valuedatetime: dt.datetime
```
haakondr, line 445: if we pass in a timezone-aware ISO8601 string, would not that resolve in a tzaware datetime object? And we don't need to worry about having a separate field for tzoffset?

src/odm2_postgres_api/schemas/schemas.py
```diff
@@ -1,13 +1,48 @@
...
+
+def create_optional_date_time_checker(datetime_name: str, offset_name: str):
+    obligatory_check = create_obligatory_date_time_checker(datetime_name, offset_name)
+
+    def check_utc_offset(date_time: dt.datetime, values):
+        if not date_time:
+            return date_time
+        if values[offset_name] is None:
+            raise ValueError(f"Missing '{offset_name}', When '{datetime_name}' is "
+                             f"supplied you must supply '{offset_name}'")
```
haakondr, line 34: maybe stick to pydantic exception types? Not sure how this is typically done

src/odm2_postgres_api/app.py
```diff
@@ -210,6 +213,80 @@ async def post_categorical_results(categorical_result: schemas.CategoricalResult
...
+                        timeaggregationinterval=0,
+                        timeaggregationintervalunitsid=18,  # time in seconds
+                        datavalue=data_value,
+                        valuedatetime=form['date'],
+                        valuedatetimeutcoffset=0
+                    )
+                    await post_measurement_results(data, connection)
+
+
 @app.post("/begroing_result", response_model=schemas.BegroingResult)
```
haakondr, line 290: I think we should type out the object structure for BegroingResult instead of just being a dict. This way requests are validated and generated openapi spec documents its structure.

in addition, the client would get proper status codes and response messages instead of internal server errors (500)

src/odm2_postgres_api/app.py
```diff
@@ -210,6 +213,80 @@ async def post_categorical_results(categorical_result: schemas.CategoricalResult
...
+        if len(used_method_indices) != 1:
+            raise ValueError('Must have one and only one method per species')
+        observations_per_method[used_method_indices[0]].append(index)
+
+    # 13 is 'Microscopic semi quantitative abundance classifier', 17 is a percentage
+    result_type_and_unit_dict = {
+        'Microscopic abundance': ("Category observation", 13, "Liquid aqueous"),
+        'Macroscopic coverage': ("Measurement", 17, "Vegetation")
+    }
+    async with connection.transaction():
```
haakondr, line 229: I think we should upload the csv within transaction, so that if csv upload fail, we don't end up with data in the database.

src/odm2_postgres_api/app.py
```diff
@@ -210,6 +213,80 @@ async def post_categorical_results(categorical_result: schemas.CategoricalResult
...
+                        censorcodecv=censor_code,
+                        qualitycodecv="None",
+                        aggregationstatisticcv="Sporadic",
+                        timeaggregationinterval=0,
+                        timeaggregationintervalunitsid=18,  # time in seconds
+                        datavalue=data_value,
+                        valuedatetime=form['date'],
+                        valuedatetimeutcoffset=0
+                    )
+                    await post_measurement_results(data, connection)
```
haakondr, line 287: I think I will need a walkthrough of this logic. A lot of things happening

src/odm2_postgres_api/utils/populate_with_general_metadata.py
```diff
@@ -5,34 +5,23 @@
...
+people_and_organizations = {"organizations": {
     "organizationtypecv": "Research institute",
     "organizationcode": "niva-no",
     "organizationname": "NIVA Norsk institut for vannforskning",
     "organizationdescription": "The Norwegian institute for water research",
     "organizationlink": "www.niva.no"
-}, "affiliations": [{
+}, "people": {
+    "personfirstname": "Uknown begroing employee",
+    "personlastname": "NIVA"
```
haakondr, line 16: as a general solution, I think we should consider to opt for "unknown person" as a general fallback, as this could be used for other data sources as well. We're not really adding much context by defining a list of "unknown <domain> employee" I think

src/odm2_postgres_api/utils/google_cloud_utils.py
```diff
@@ -28,19 +28,23 @@ def set_uploader(self):
 
 
 def generate_csv_from_form(form):
+    observation_date = dt.datetime.fromisoformat(form['date'].replace('Z', '+00:00'))
+    date_string = (observation_date + dt.timedelta(hours=12)).date().strftime('%d-%m-%Y %H:%M:%S')  # round date
```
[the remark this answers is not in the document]
RmStorm: It's rounding to the nearest date for creating the csv file.

src/odm2_postgres_api/app.py
RmStorm: completely agree

src/odm2_postgres_api/app.py
RmStorm: that's clever

src/odm2_postgres_api/schemas/schemas.py
RmStorm: yes, there is a choice to be made here. Right now the schemas mirror the database. Since the schemas are used/populated by fastapi those fields have to be submitted by the client.

The issue here is that pydantic doesn't differentiate between datetimes with and without timezone while postgres does. This means we need to manually enforce timezoneless datetimes to be compatible with postgres. I don't want to just strip the timezone without validating that that info is properly retained in the utcoffset field. Ideally I would set that field but that's not possible as far as I can tell.

src/odm2_postgres_api/schemas/schemas.py
RmStorm: Pydantic catches this and uses it to formulate it's errors. I copied this structure from the pydantic documentation:
https://pydantic-docs.helpmanual.io/usage/validators/

src/odm2_postgres_api/app.py
RmStorm: yup, It's a direct copy of the code I wrote in discussion with Jens to figure out how to make the semi quantitative begroing data best conform to the ODM2 data model. I use something like this to copy the data points from begroing into ODM2. I guess that code can be simplified now that this logic resides here in the api.

RmStorm:
@haakondr I think I will merge this now so it can be tested tomorrow. There are two issues remaining that we need to decide on:
- How best to deal with separate date and utcoffset fields
- How to deal with unknown persons

And we need to go through the observation insertion code together.

src/odm2_postgres_api/schemas/schemas.py
haakondr: hm. while pydantic does not differentiate datetimes, can we write a validation that enforces the client to set timezone (tz), such that deserialized datetimes always are timezone aware?

And - how does our database write behave in cases of timezone aware datetimes? do they get written as tz-aware timestamps/time values?

haakondr: If I understand you correctly, the main issue is that ODM2 schema has two fields, while in postgres it would be possible to write this as one tz-aware column?

I suggest we introduce some mapping serverside so that our API remains clean. I think the API structure is more important than internal implementation details. Technical depth that propagates into multiple systems are typically harder to deal with than some inline hack within a API. 

## Alternative solutions

- one solution would be to manually handle this for the tables in question. We have our "poor-maps ORM" solution "insert_pydantic_object".  
- Looking at results.resultdatetimeutcoffset, this field is not required. How about we omit writing to this, and just write timezone-aware dates into resultdatetime? This would diverge from the ODM2 schema, but probably result in the cleanest code.

what do you think?

PR #34, opened 2020-07-10
Add metadata for begroing indices, closes #24

### LINKED ISSUE #24: add metadata for begroing indices
figure out what the metadata for the begroing indices should look like.

RmStorm:


src/odm2_postgres_api/utils/populate_with_general_metadata.py
```diff
@@ -267,19 +357,29 @@
...
 
 
 def post_to_odm2_api(endpoint, data):
     response = requests.post('http://localhost:8701/' + endpoint, json=data)
-    response.raise_for_status()
+    try:
+        response.raise_for_status()
+    except Exception:
+        logging.error(data)
+        raise
```
haakondr, line 375: how does this differ from raise_from_status?

RmStorm: It logs the data with which the request was made instead of only the reply.

Issue #43, opened 2020-08-10
Import of on-prem users cause duplicates


haakondr:
Given the following sequence:

1. import users from on-prem AD, see issue #38 
2. log in with "employee login" with azure AD in https://test.niva.no (test) or https://data.niva.no (prod)
3. save something from begroing app at https://begroing.test.niva.no/

This should trigger the logic in user.py create_or_get_user. Here, we look for existing users in personexternalidentifiers, but this does not exist on first time interaction, even though the person exists in people table (from step 1). This results in duplicate people rows for the same user:

```
select p.*, a.primaryemail
from odm2.people p inner join odm2.affiliations a on p.personid=a.personid where p.personfirstname='Håkon';
```

| personid | personfirstname | personmiddlename | personlastname | primaryemail |
| :--- | :--- | :--- | :--- | :--- |
| 31 | Håkon | Drolsum | Røkenes | Hakon.Drolsum.Rokenes@niva.no |
| 252 | Håkon | Drolsum | Røkenes | Hakon.Drolsum.Rokenes@niva.no |

## suggested solution

- we treat the users email as the unique identifier instead of port user id
- drop all references to port user id. This is internal to niva port and may lead to inconsistencies. It is probably better to make odm2 db unaware of where the user came from (?)

## side notes

- we do these queries to fetch user for each request to odm2 API /begroing_result. This is not ideal, and we should really fetch this information from the JWT tokens. This is currently not setup, but we should consider integrating niva-port more tightly with the odm2 people table?

haakondr:
Merged to dev. Tested by doing the following:

- cleaned up previous mess by deleting my own user in odm2 tables people, affiliations, actionby and personexternalidentifiers in nivatest
- logged in, submitted a result in begroing app

I now got only one people row:

```
select * from odm2.people where personfirstname = 'Håkon';
```

| personid | personfirstname | personmiddlename | personlastname |
| :--- | :--- | :--- | :--- |
| 31 | Håkon | Drolsum | Røkenes |

Also, the result was mapped to my user:

```
select p.personfirstname, a.primaryemail, ac.actiontypecv, pe.personexternalidentifier
from odm2.people p
         inner join odm2.affiliations a
         inner join odm2.actionby ab on ab.affiliationid = a.affiliationid on p.personid = a.personid
        inner join odm2.actions ac on ac.actionid=ab.actionid
        inner join odm2.personexternalidentifiers pe on pe.personid=p.personid
[...]

PR #38, opened 2020-07-29
Wrote script to manually import users from csv file exported from on-prem AD #64

haakondr:
1. exported using powershell connected to niva network on my windows machine:

```
Get-ADUser -Filter * -SearchBase "OU=OSLO,OU=NIVA,DC=niva,DC=corp" |
Select UserPrincipalName, GivenName, Surname, SamAccountName |
Export-CSV -Delimiter ";" -Encoding UTF8 -Path
"c:\niva\port\utils\data\emails.csv"
```

2. sent the file to my linux machine
3. on linux  I ran the following to fix encodings:

```
dos2unix data/emails.csv
```

4. deleted first line as it is not valid, then ran script

## features:

- duplicate detection based on person name. If person exists, we circuit
break and stop importing
- if user does not have first or last name, we ignore user (some system
users etc, we probably don't want these in the system)

## TODO:

- validate that external identifier does not already exist (as in, have
we already imported person with SamAccountName?). Currently we just
check person name, but people can change names so that is probably not
100% solid
- some fixing up, like setting connections etc based on env variables. Planning on running in nivatest when PR is approved, will fix that then on friday

PR #45, opened 2020-08-10
Changed logic for getting stored people/affiliations #43

haakondr:
Fetching stored person based on user email instead of port-id. This
solves issues when people row was populated from on-prem active
directory export (see #38)

Also removed the concept of external identifier to niva-port. This makes
odm2 unaware of whether the person was populated from a logged in user
or other sources. This may be a drawback of this approach, but I think
it also makes the system more robust by just referring to users with
their UserPrincipalName (email)

changes:

- removed niva-port as externalidentifiersystem in metadata script
- fetch user based on affiliations.primaryemail instead of
externalidentifiersystem


haakondr:
@RmStorm since you are on vacation, I took the liberty to merge this to dev, but made a PR FYI

PR #46, opened 2020-08-10
added unique constraint on affiliations.primaryemail

haakondr:
this is mainly to prevent duplicate people entries, see issue #43

Question is, can users have multiple affiliations with the same email address? and is email address a proper unique identifier for a user?


haakondr:
Closing PR as roald is on vacation, PR nagging on slack is annoying. Lets pick this up when you are back :)

RmStorm (approved):
This is a good change. As far as I can see there are no problems with it. I was thinking that maybe affiliations can be used to encode different roles a single person might fulfill in a single organization thus leading to conflicts but that doesn't seem to be the case. So, LGTM

haakondr:
> This is a good change. As far as I can see there are no problems with it. I was thinking that maybe affiliations can be used to encode different roles a single person might fulfill in a single organization thus leading to conflicts but that doesn't seem to be the case. So, LGTM

Could be the case that a user has multiple affiliations. Not sure. If that is the case, I guess we can just remove the unique constraint.

PR #52, opened 2020-08-24
Move routes to separate directory and isolate pool_manager

roryscot:
Refactor behemoth list of routes into modules. 

NOTE: we are exporting a single instance of 'api_pool_manager' from 'src/odm2_postgres_api/utils/api_pool_manager.py'. Are there any obvious problems with this approach (e.g. is it fragile/ bad practice)

### DIFF
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -26,3 +26,12 @@ async def truncate_all_data(connection, schema: str):
     rows = await connection.fetch("select * from pg_tables where schemaname=$1;", schema)
     tables = [t["tablename"] for t in rows]
     await _truncate_tables(schema=schema, tables=tables)
+
+
+def test_pool_mamager_uses_same_instance():
+    from odm2_postgres_api.routes.shared import shared_routes
+    from odm2_postgres_api.routes.begroing_routes import begroing_routes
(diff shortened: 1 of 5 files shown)


tests/test_utils.py
```diff
@@ -26,3 +26,12 @@ async def _truncate_tables(schema, tables, iterations=10):
     rows = await connection.fetch("select * from pg_tables where schemaname=$1;", schema)
     tables = [t["tablename"] for t in rows]
     await _truncate_tables(schema=schema, tables=tables)
+
+
+def test_pool_mamager_uses_same_instance():
```
RmStorm, line 31: mamager typo!

RmStorm (approved):
Looks good to me, there is a little typo in the test though.

haakondr (approved):
we should perhaps consider to organise our paths in the same way? 

- the stuff in begroing_routes.py behind POST /begroing/observation (or something like that)
- shared behind /shared/*

Also, I'd prefer to not have folders if we only stick one file in it, could we do something like:

routes/begroing_routes.py
routes/shared_routes.py

PS: don't think we need to rename paths now as that would involve also changing the frontend, just throwing it out there

PR #54, opened 2020-08-28
db_init: waiting for database to be ready before running

haakondr:
- instead of doing restart via docker-compose we have our script to wait
until db is up and ready. This should be more robust, as doing restart
could lead to script restarting in a infinite loop


docker-compose.yml
```diff
@@ -10,7 +10,6 @@ services:
     depends_on:
       - timescale_odm2
       - db_init
-    restart: always
```
haakondr, line 13: should probably add something equivalent in the API's as well?

Using restart: always is not very good. At least on my machine I very often have various containers in infinite restart loops

RmStorm (approved):
Looks good! I ran it locally and it works. I think something similar indeed has to be done in the api but this can be merged first.

PR #55, opened 2020-09-01
Feature/create indices

roryscot:
Depends on https://github.com/NIVANorge/begroing-data-upload/issues/63


src/odm2_postgres_api/routes/begroing_routes.py
```diff
@@ -99,3 +113,56 @@ async def post_begroing_result(begroing_result: schemas.BegroingResultCreate,
...
+            qualitycodecv="None",
+            aggregationstatisticcv="Unknown",
+            timeaggregationinterval=0,
+            timeaggregationintervalunitsid=18,  # time in seconds
+            datavalue=index_instance.indexValue,
+            valuedatetime=new_index.date,
+            valuedatetimeutcoffset=0
+        )
+
+        await post_measurement_results(data_measurement_result, connection)
```
haakondr, line 166: reusing the http methods here. There is probably nothing hindering from using them directly, but I think that we should strive to call them only over http to have a unified way in.

I suggest we either call core_queries.upsert_measurement_result directly or extract the functionality out to a shared function (currently it is only contains one code-line)

src/odm2_postgres_api/routes/begroing_routes.py
```diff
@@ -99,3 +113,56 @@ async def post_begroing_result(begroing_result: schemas.BegroingResultCreate,
...
+
+    return schemas.BegroingIndices(personid=user.personid, **new_index.dict())
```
haakondr, line 168: Ideally I think we should return the object structure that was written to the database. In this case we try to reconstruct the same structure. If we return the returned objects from the sql queries then we can also add unit tests that test the response.

In this specific case I would improve upsert_measurement_result and return the written rows and build the BegroingIndices object

src/odm2_postgres_api/routes/begroing_routes.py
```diff
@@ -99,3 +113,56 @@ async def post_begroing_result(begroing_result: schemas.BegroingResultCreate,
...
+    )
+
+    completed_action = await post_actions(data_action, connection)
+
+    for index_instance in new_index.indices:
+        data_result = schemas.ResultsCreate(
+            samplingfeatureuuid=new_index.station['samplingfeatureuuid'],
+            actionid=completed_action.actionid,
+            resultuuid=str(uuid.uuid4()),
+            resulttypecv="Measurement",
```
haakondr, line 142: note to self: would be great to publish a pypi package with ODM2 CV's as enums

src/odm2_postgres_api/routes/begroing_routes.py
```diff
@@ -99,3 +113,56 @@ async def post_begroing_result(begroing_result: schemas.BegroingResultCreate,
...
+            variableid=INDEX_NAME_TO_VARIABLE_ID[index_instance.indexType],
+            unitsid=19,  # Dimensionless,-,Dimensionless
+            processinglevelid=1,  # id:1, "processinglevelcode": "0", "definition": "Raw Data"
```
haakondr, line 145: Maybe we should not refer to processinglevels as id but rather fetch processinglevelcode = '0'. I got confused here first and thought we were inserting code = 1, but we are really inserting id=1 which is processinglevelcode=0 (did not read comment, usually don't read them).

Problem with this approach is that it is very tightly coupled to metadata setup. What happens if we initialize a new database and forget to reset id sequences. Then the application breaks.

I think we need to fetch the processinglevelid for a given code and never refer to internal id's in our application. This is where ODM2 normalized structure has gone a bit too far IMO as this is quite cumbersome to work with.

Same goes for units above.

@RmStorm thoughts?

src/odm2_postgres_api/routes/begroing_routes.py
```diff
@@ -99,3 +113,56 @@ async def post_begroing_result(begroing_result: schemas.BegroingResultCreate,
...
+        begindatetimeutcoffset=0,
+        equipmentids=[],
+        directiveids=[e.directiveid for e in new_index.projects]
+    )
+
+    completed_action = await post_actions(data_action, connection)
+
+    for index_instance in new_index.indices:
+        data_result = schemas.ResultsCreate(
+            samplingfeatureuuid=new_index.station['samplingfeatureuuid'],
```
haakondr, line 139: we should create a type definition for station

src/odm2_postgres_api/routes/begroing_routes.py
```diff
@@ -99,3 +113,56 @@ async def post_begroing_result(begroing_result: schemas.BegroingResultCreate,
...
+                       connection=Depends(api_pool_manager.get_conn),
+                       niva_user: str = Header(None)):
+    user = await create_or_get_user(connection, niva_user)
+
+    data_action = schemas.ActionsCreate(
+        affiliationid=user.affiliationid,
+        isactionlead=True,
+        methodcode="begroing_6",  # code begroing_6 = Begroing Index Calculation
+        actiontypecv="Derivation",
+        begindatetime=new_index.date,
```
haakondr, line 129: I don't think we need to improve this now, but:

The UI posts dates in the following format:

```
2020-09-02T00:00:00.000Z
```

This contains timezone information (Z being UTC). The date coming in is in this case a timezone-aware date and we might as well just use the incoming datetime object to derive the utcoffset. I see that @RmStorm has created validations to check if specified offset matches actual offset, but maybe we can just derive all this?

this way we'd have less tightly coupling with the clientside. With current implementation the client HAS to use utc, and that is not documented anywhere

src/odm2_postgres_api/routes/begroing_routes.py
```diff
@@ -99,3 +113,56 @@ async def post_begroing_result(begroing_result: schemas.BegroingResultCreate,
...
+        begindatetimeutcoffset=0,
+        equipmentids=[],
```
haakondr, line 131: are there any equipment used that should be stored within the ODM2 database? if yes, I think we should create issues in github for this

src/odm2_postgres_api/routes/begroing_routes.py
```diff
@@ -99,3 +113,56 @@ async def post_begroing_result(begroing_result: schemas.BegroingResultCreate,
...
+        data_measurement_result = schemas.MeasurementResultsCreate(
+            resultid=completed_result.resultid,
+            censorcodecv="Not censored",
+            qualitycodecv="None",
+            aggregationstatisticcv="Unknown",
+            timeaggregationinterval=0,
+            timeaggregationintervalunitsid=18,  # time in seconds
+            datavalue=index_instance.indexValue,
+            valuedatetime=new_index.date,
+            valuedatetimeutcoffset=0
```
haakondr, line 163: What does the data structure for a index look like?

is it 1 value per begroing observation? Or an aggregate over time? Should be documented in https://github.com/NIVANorge/begroing-data-upload/issues/98

src/odm2_postgres_api/queries/core_queries.py
```diff
@@ -240,3 +242,14 @@ async def upsert_categorical_result(conn: asyncpg.connection, categorical_result
...
+
+
+async def get_index_names(conn: asyncpg.connection):
+    async with conn.transaction():
+        # value_keys = []
+
+        # await conn.fetchrow(make_sql_query('categoricalresults', categoricalresults_data),
+        #                     *categoricalresults_data.values())
+        # await conn.fetchrow(make_sql_query('categoricalresultvalues', categoricalresultvalues_data),
+        #                     *categoricalresultvalues_data.values())
```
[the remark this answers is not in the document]
haakondr: Also see that this is unused. Will it be used or can it be deleted?

tests/test_api.py
```diff
@@ -58,3 +58,26 @@ def test_create_units(wait_for_db, clear_db):
...
+            "samplingfeaturename":"AAL ALN1, Elveovervåking, Alna, St.1 (RID)",
+            "samplingfeaturedescription":"Station from begroingsdatabase,\noriginal projection: 32632",
+            "featuregeometrywkt":"POINT (10.791888159215542 59.90455564858938)"
+        },
+        "date":"2020-09-01T00:00:00.000Z",
+        "indices":[{"indexType":"PIT EQR","indexValue":"11"}]
+    }
+    headers = {
+        "Content-Type": "application/json",
+        "Niva-User": "eyJpZCI6IjEiLCJlbWFpbCI6ImJlZ3JvaW5nQGRldnVzZXIuY2…kaXNwbGF5TmFtZSI6IkFsZnJlZCBCZWdyb2luZyBPbHNlbiJ9"
```
haakondr, line 78: for readability it is probably better to create a NivaPortUser object and base64 encode it:

```
b64encode(json.dumps(user_obj).encode('utf-8'))
```

We could look into adding some util functions to do this in the future.

tests/test_api.py
```diff
@@ -58,3 +58,26 @@ def test_create_units(wait_for_db, clear_db):
...
+    }
+
+    with TestClient(app) as client:
+        response = client.post(url="/indices", data=json.dumps(index_data), headers=headers)
```
haakondr, line 82: not really testing anything here. I think we need to assert the response that all values look ok. This relates to my comment about returning written values in the endpoint. 

If the endpoint returns the written values from sql as response we can assert on it here in the tests

src/odm2_postgres_api/routes/begroing_routes.py
```diff
@@ -99,3 +113,56 @@ async def post_begroing_result(begroing_result: schemas.BegroingResultCreate,
     # TODO: Send email about new bucket_files
 
     return schemas.BegroingResult(personid=user.personid, **begroing_result.dict())
+
+
+@router.post("/indices", response_model=schemas.BegroingIndices)
+async def post_indices(new_index: schemas.BegroingIndicesCreate,
+                       connection=Depends(api_pool_manager.get_conn),
+                       niva_user: str = Header(None)):
```
haakondr, line 121: Could use a docstring here describing what the endpoint does. From what I've understood:

``` 
"""
Stores a list of indices for observations on a given timestamp/station. One observation may have many indices.
"""
```

haakondr (changes requested):
made a lot of nitpicky comments. Don't think we need to fix all of these, but I think the following needs to be fixed:

- type definition for stations/samplingfeatures
- remove dead code (get_index_names) if not used
- different http status code if method does not exist
- consider to refer to processinglevelcode etc by code instead of id
- improve endpoint so that it responds written values and assert this in the API test

src/odm2_postgres_api/queries/core_queries.py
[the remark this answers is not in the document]
RmStorm: hehe we were trying to get the number 30 included in a doc string so `pytest --fixtures` would give a nice informative doc string. It didn't work and this is a leftover from that attempt :laughing:

tests/integration_test_fixtures.py
```diff
@@ -10,8 +10,8 @@
 from odm2_postgres_api.utils.db_initiate import db_init
 from test_utils import truncate_all_data
 
-
-def try_db_init(tries=30):
+DEFAULT_TRIES = 30
+def try_db_init(tries=DEFAULT_TRIES):
```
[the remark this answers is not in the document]
RmStorm: hehe we were trying to get the number 30 included in a doc string so `pytest --fixtures` would give a nice informative doc string. It didn't work and this is a leftover from that attempt :laughing:

src/odm2_postgres_api/queries/core_queries.py
```diff
@@ -76,6 +76,8 @@ async def insert_method(conn: asyncpg.connection, method: schemas.MethodsCreate)
 async def do_action(conn: asyncpg.connection, action: schemas.ActionsCreate):
     async with conn.transaction():
         method_row = await conn.fetchrow("SELECT methodid FROM methods WHERE methodcode = $1", action.methodcode)
+        if method_row is None:
+            raise HTTPException(status_code=404, detail="Method not found.")
```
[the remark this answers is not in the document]
RmStorm: agreed!

src/odm2_postgres_api/routes/begroing_routes.py
RmStorm: Agreed when there is a unique code in a table we should use that instead of id!

src/odm2_postgres_api/routes/begroing_routes.py
RmStorm: hmm, I think aggregate over time.

src/odm2_postgres_api/routes/begroing_routes.py
RmStorm: yes, I think that can make sense. I didn't do it for the observations because I wanted to have something working. I think you're right though that both of these methods should return the actual database entries...

tests/integration_test_fixtures.py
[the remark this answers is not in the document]
haakondr: aha, I understand

src/odm2_postgres_api/queries/core_queries.py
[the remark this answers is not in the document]
roryscot: raise HTTPException(status_code=422, detail="Please specify valid methodcode.")
?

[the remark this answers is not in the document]
haakondr: yep, that's fine. I am not sure whether this results in a json response or a text/html response. We'd probably want to use json response.

This does depend on whether there is a exceptionhandler for HTTPException that responds with json

src/odm2_postgres_api/routes/begroing_routes.py
```diff
@@ -99,3 +113,56 @@ async def post_begroing_result(begroing_result: schemas.BegroingResultCreate,
     # TODO: Send email about new bucket_files
 
     return schemas.BegroingResult(personid=user.personid, **begroing_result.dict())
+
+
+@router.post("/indices", response_model=schemas.BegroingIndices)
```
haakondr, line 118: Since we split out begroing_routes into its own file, we should probably also reflect this in the url structure. I propose the following:

- POST /begroing/observations
- POST /begroing/indices

src/odm2_postgres_api/queries/core_queries.py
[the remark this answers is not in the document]
roryscot: @RmStorm ?

src/odm2_postgres_api/routes/begroing_routes.py
haakondr, line 118: this endpoint is not transactional, as far as I can tell. This is important. Basically we'd want to wrap all this code with

```
```

This way all writes are rolled back if something goes wrong

src/odm2_postgres_api/schemas/schemas.py
```diff
@@ -510,6 +510,22 @@ class BegroingResult(BegroingResultCreate):
     personid: int
 
 
+class IndicesInfo(BaseModel):
+    indexType: str
+    indexValue: float
+
+
+class BegroingIndicesCreate(BaseModel):
+    projects: List[Directive]
```
haakondr, line 519: I think we should validate that we have at least one item in our list. Currently empty list is valid.

It seems like this can be done with "conlist":

https://stackoverflow.com/questions/61468548/check-if-list-is-not-empty-with-pydantic-in-an-elegant-way

```
projects: conlist(Directive, min_items=1)
```

Have not tried this, just reading docs

src/odm2_postgres_api/routes/begroing_routes.py
roryscot: Should we also do that with line 33: post_begroing_result ?

haakondr: Yes

RmStorm: post begroing result already has this!

roryscot:
Did we want to do anything else before merging?

PR #57, opened 2020-09-03
Feature/mass spectrometry

ZofiaRudjord:
Specify a minimal set of metadata necessary to run mass spec analysis using feature/odm2 branch of nivaloud-mass_spectrometry.
Naming convention for get function should be discussed and changed accordingly.


RmStorm (approved):
We have gone through the structure in quite a lot of detail (routing and metadata creation etc.). But I have not taken a _close_ look at the new queries. Since I have tested the other masspec code that uses these queries I think it should be ok. When we merge the new odm2 code in the massspec repository we should be sure to write integration tests that hit these new endpoints.

PR #59, opened 2020-09-07
49 metadata setup

haakondr:
## core changes
- running metadata setup as part of app startup. Moved away from using http to executing the functions directly. Main motivation for doing this was that it was not straight forward to wait for http server to be up and ready (fastapi lifecycle event "startup" is before starting). I think this is more solid anyway, no reason to do this over http

- refactored the script so that sequence of execution is a bit more
readable. Slit metadata setup into different files
- ~~trigger CV download as part of script~~
  - this has been changed in #60, please have a look at that after
- added checks in functions if row exists, we do nothing. This makes the script idempotent, but does not support updates. May consider to support this in the future
- dynamically setting niva org id in cases where it was hardcoded
  - wrap all metadata definitions as functions. This is used to pass in
props like niva org id, but do this for all to ensure consistency. Maybe
unnecessary (?) 
- moved begroing specific stuff to its own file
- added some unique constraints to avoid duplicates

## misc changes

- moved db_init and config to its own folder
- differentiated cv create and response (cv create has CV_TABLE as part
of it). Could consider to move this to pathparam as well?

## TODOS

-no way to update org, person, etc. as it is now. Only prevent
duplicates
- begroing method annotations set to [1]. What does that mean?


src/db_init/ODM2_for_PostgreSQL.sql
```diff
@@ -2326,4 +2326,7 @@ ALTER TABLE ODM2.Annotations ADD COLUMN annotationjson json;
 ALTER TABLE ODM2.personexternalidentifiers
     ADD UNIQUE (personexternalidentifier, externalidentifiersystemid);
 
-ALTER TABLE ODM2.affiliations ADD UNIQUE (primaryemail);
\ No newline at end of file
+ALTER TABLE ODM2.affiliations ADD UNIQUE (primaryemail);
+
+ALTER TABLE ODM2.externalidentifiersystems ADD UNIQUE (externalidentifiersystemname);
+ALTER TABLE ODM2.units ADD UNIQUE (unitstypecv, unitsabbreviation);
```
haakondr, line 2332: there are probably more unique constraints that we can consider adding

src/odm2_postgres_api/app.py
```diff
@@ -44,9 +45,12 @@ async def startup_event():
 
     logging.info("Creating connection pool")
     api_pool_manager.pool = await asyncpg.create_pool(user=db_mighty_user, password=db_mighty_pwd,
+                                                      server_settings={"search_path": "odm2"},
                                                       host=db_host, port=db_port, database=db_name)
     logging.info("Successfully created connection pool")
 
+    await populate_metadata(api_pool_manager.pool)
```
haakondr, line 52: this is the big change, running this on app startup. This makes local dev much more pleasant, and is also automatically ran in test/prod. 

Does add some to app startup, but that is taken care of by rolling deployment.

downside to this approach is that we run this once per replica, so if we have many replicas there will be some load on our database during deployment but  I don't think this should be an issue. The setup should now be idempotent (I hope) so it can be ran many times

src/odm2_postgres_api/controlled_vocabularies/cv_definitions/cv_actiontype.json
```diff
@@ -0,0 +1,308 @@
+{
```
haakondr, line 1: ran my new script download.csv.py and stored results in this folder. This makes the workflow of populating with CV's much easier

haakondr:
```
 du cv_definitions/ -sh
1020K	cv_definitions/
```

1mb of files. Not too much so I think this fits into git. 

I prettied them for readability but we can consider to not do that to minimize file size

src/odm2_postgres_api/controlled_vocabularies/download_cvs.py
```diff
@@ -0,0 +1,71 @@
...
+
+async def fetch(session, url):
+    logging.info("Downloading CV", extra={"url": url})
+    async with session.get(url) as response:
+        if response.status != 200:
+            response.raise_for_status()
+        return await response.text()
+
+
+CV_URL = "http://vocabulary.odm2.org/api/v1/%s/?format=json"
```
haakondr, line 48: notice the undocumented "format=json" option

src/odm2_postgres_api/controlled_vocabularies/load_cvs.py
```diff
@@ -0,0 +1,27 @@
...
+    path = os.path.join(os.path.dirname(__file__), "cv_definitions", f"{cv_name}.json")
+    logging.info("Loading CV's from file", extra={"file": path})
+    with open(path) as f:
+        o = json.load(f)
+        return [ControlledVocabulary(**cv) for cv in o["objects"]]
+
+
+async def load_controlled_vocabularies(db_pool):
+    logging.info("Loading controlled vocabularies defined as json")
+    async with db_pool.acquire() as conn:
```
haakondr, line 21: acquiring just one connection and run things in sequence. Could perhaps be speed up by doing things async but did not bother rewriting

src/odm2_postgres_api/metadata_init/populate_metadata.py
```diff
@@ -0,0 +1,57 @@
...
+        for cv in begroing_controlled_vocabularies():
+            await save_controlled_vocab(conn, cv)
+
+        for v in begroing_variables():
+            await save_variables(conn, v)
+
+        for m in begroing_methods(org_id=niva_org_id):
+            await save_methods(conn, m)
+
+    logging.info("ODM2 metadata init done")
```
haakondr, line 57: this is the main script for populating metadata. Could be rewritten to async by using the db pool, but don't think there is a lot to save

src/odm2_postgres_api/metadata_init/repository.py
```diff
@@ -0,0 +1,80 @@
...
+    if existing:
+        return existing
+    return await insert_pydantic_object(connection, 'variables', variable, Variables)
+
+
+async def save_methods(connection, method: MethodsCreate) -> Methods:
+    existing = await find_row(connection, "methods", "methodcode", method.methodcode, Methods)
+    if existing:
+        return existing
+    return await core_queries.insert_method(connection, method)
```
haakondr, line 80: took some of the API endpoints out. I don't think any of the API endpoints in shared_routes.py can be used in their current state so may consider to just remove them all?

src/odm2_postgres_api/controlled_vocabularies/cv_definitions/cv_actiontype.json
RmStorm: prettifying adds minimal storage space and adds a ton of readability so having them be pretty is definitely the way to go.

src/odm2_postgres_api/controlled_vocabularies/download_cvs.py
```diff
@@ -0,0 +1,71 @@
...
+    logging.info("Downloading controlled vocabularies", extra={"url": CV_URL})
+    async with aiohttp.ClientSession() as session:
+        tasks = [download_and_save_cv(session, cv_name) for cv_name in table_names]
+        responses = await asyncio.wait(tasks)
+    logging.info("Done downloading controlled vocabularies. If there are any git diffs, check these in")
+
+
+if __name__ == '__main__':
+    setup_logging(plaintext=True)
+    loop = asyncio.get_event_loop()
```
RmStorm, line 70: I don't think you need to explicitly get the loop. The preferred way to do this since (3.6? 3.7?) is to simply call 
```asyncio.run(main())```

File: src/odm2_postgres_api/controlled_vocabularies/load_cvs.py [hunk 9]
```diff
@@ -0,0 +1,27 @@
...
+        for cv_name in CONTROLLED_VOCABULARY_TABLE_NAMES:
+            cvs = read_cv_from_file(cv_name)
```
RmStorm, line 23: maybe `cvs` here can be renamed to `cv_entries`?

src/odm2_postgres_api/main.py
```diff
@@ -13,9 +13,9 @@
     port = 5000
     if os.environ.get('NIVA_ENVIRONMENT') not in ['dev', 'master']:
         if Path.cwd() == Path('/app'):
-            env_file = Path(__file__).parent / 'config' / 'localdocker.env'
+            env_file = Path(__file__).parent.parent / 'config' / 'localdocker.env'
```
RmStorm, line 16: `Path(__file__).parent.parent` == `Path(__file__).parents[1]`
The second notation has my prefence :sweat_smile:

src/odm2_postgres_api/queries/core_queries.py
```diff
@@ -21,11 +23,30 @@ def make_sql_query(table: str, data: dict):
 
 
 async def insert_pydantic_object(conn: asyncpg.connection, table_name: str, pydantic_object, response_model):
+    logging.info(f"Inserting row", extra={"table": table_name})
```
RmStorm, line 26: maybe .debug?

src/odm2_postgres_api/queries/core_queries.py
```diff
@@ -21,11 +23,30 @@ def make_sql_query(table: str, data: dict):
...
+                              unit.unitsabbreviation)
+    return Units(**row) if row else None
+
+
+async def find_row(conn: asyncpg.connection, table: str, id_column: str, identifier, model):
+    row = await conn.fetchrow(f"SELECT * FROM {table} WHERE {id_column}=$1", identifier)
+    return model(**row) if row else None
+
+
+async def find_organization(conn: asyncpg.connection, org_code: str) -> Optional[Organizations]:
```
RmStorm, line 43: This one fits in find_row right?

haakondr: hmm, yes. Think I wrote it before I wrote generic function find_row

src/odm2_postgres_api/controlled_vocabularies/download_cvs.py
haakondr: yep that worked. thanks

src/odm2_postgres_api/main.py
haakondr: moved config back into its original location so this is no longer needed

src/odm2_postgres_api/queries/core_queries.py
haakondr: I see that this one is actually not in use anymore. I wrote it initially, then made find_row and forgot to delete this

haakondr:
made a commit with PR feedback, should be all done now I think

RmStorm (changes requested):
I think this is in principle good to go there is just two things.

1. The tests should pass. we can probably just set sensible defaults in the the integration test fixtures.
2. I'm not sure we should remove all these endpoints in `shared_routes.py` I agree that in the future they *might* be replaced by something else but for now they connect to the queries and I think should just be left in place? We're going to have to build some sort of interface for managing that kind of metadata and then we will have to recreate these endpoints?

RmStorm:
Also, can we rename the file `repository.py` to something more informative? There is a bunch of functions in there that are basically idempotent wrappers for `core_queries.py` as far as I can tell. I don't really have a good suggestion though..  `idempotent_core_helpers.py` maybe? :rofl:

haakondr:
> core_queries

renamed it "storage". What do you think?

```
from storage import save_organisation
```

RmStorm (approved):
Looks great. Makes it much more usable. We should keep in mind to re-evaluate this in the future once the metadata becomes fully managed by the database.

PR #61, opened 2020-09-08
Expanded the FeatureActionsCreate schema to be more general

RmStorm:
A feature_action row requires an actionid and a samplingfeatureid. Since
it proved cumbersome to keep track of samplingfeatureid's the schema
allowed for uuids instead and finds the associated samplingfeature.

Since both `samplingfeatureuuid` and `samplingfeaturecode` have a unique
constraint it makes sense to allow either when trying to find a
samplingfeatureid that should be referenced in a feature_action row. If
both a uuid and a code are supplied they have to both match to be
valid. If only one is supplied that one is simply used to query for a
samplingfeature.

### DIFF
--- a/src/odm2_postgres_api/queries/core_queries.py
+++ b/src/odm2_postgres_api/queries/core_queries.py
@@ -147,18 +147,30 @@ async def create_result_data_quality(conn: asyncpg.connection, result_data_quali
 
 
 async def create_feature_action(conn: asyncpg.connection, feature_action: schemas.FeatureActionsCreate):
-    row = await conn.fetchrow(
-        "INSERT INTO featureactions (samplingfeatureid, actionid) "
-        "VALUES ((SELECT samplingfeatureid FROM samplingfeatures where samplingfeatureuuid = $1), $2) "
-        "ON CONFLICT (samplingfeatureid, actionid) DO UPDATE SET actionid = EXCLUDED.actionid returning *",
-        feature_action.samplingfeatureuuid, feature_action.actionid)
-    return schemas.FeatureActions(samplingfeatureuuid=feature_action.samplingfeatureuuid, **row)
+    if feature_action.samplingfeatureuuid and feature_action.samplingfeaturecode:
+        row = await conn.fetchrow(
+            "INSERT INTO featureactions (samplingfeatureid, actionid) "
+            "VALUES ((SELECT samplingfeatureid FROM samplingfeatures "
+            "where samplingfeatureuuid = $1 AND samplingfeaturecode = $2), $3) "
+            "ON CONFLICT (samplingfeatureid, actionid) DO UPDATE SET actionid = EXCLUDED.actionid returning *",
+            feature_action.samplingfeatureuuid, feature_action.samplingfeaturecode, feature_action.actionid)
+        return schemas.FeatureActions(samplingfeatureuuid=feature_action.samplingfeatureuuid, **row)
+    else:
+        row = await conn.fetchrow(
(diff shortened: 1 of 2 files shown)


src/odm2_postgres_api/queries/core_queries.py
```diff
@@ -147,18 +147,30 @@ async def create_result_data_quality(conn: asyncpg.connection, result_data_quali
...
+        row = await conn.fetchrow(
+            "INSERT INTO featureactions (samplingfeatureid, actionid) "
+            "VALUES ((SELECT samplingfeatureid FROM samplingfeatures "
+            "where samplingfeatureuuid = $1 OR samplingfeaturecode = $2), $3) "
+            "ON CONFLICT (samplingfeatureid, actionid) DO UPDATE SET actionid = EXCLUDED.actionid returning *",
+            feature_action.samplingfeatureuuid, feature_action.samplingfeaturecode, feature_action.actionid)
+        return schemas.FeatureActions(samplingfeatureuuid=feature_action.samplingfeatureuuid, **row)
 
 
 async def create_result(conn: asyncpg.connection, result: schemas.ResultsCreate):
```
haakondr, line 168: should we just expose FeatureActionsCreate to the function signature?

src/odm2_postgres_api/queries/core_queries.py
```diff
@@ -147,18 +147,30 @@ async def create_result_data_quality(conn: asyncpg.connection, result_data_quali
...
+            feature_action.samplingfeatureuuid, feature_action.samplingfeaturecode, feature_action.actionid)
+        return schemas.FeatureActions(samplingfeatureuuid=feature_action.samplingfeatureuuid, **row)
+    else:
+        row = await conn.fetchrow(
+            "INSERT INTO featureactions (samplingfeatureid, actionid) "
+            "VALUES ((SELECT samplingfeatureid FROM samplingfeatures "
+            "where samplingfeatureuuid = $1 OR samplingfeaturecode = $2), $3) "
+            "ON CONFLICT (samplingfeatureid, actionid) DO UPDATE SET actionid = EXCLUDED.actionid returning *",
+            feature_action.samplingfeatureuuid, feature_action.samplingfeaturecode, feature_action.actionid)
+        return schemas.FeatureActions(samplingfeatureuuid=feature_action.samplingfeatureuuid, **row)
```
haakondr, line 165: maybe consider to have two different functions. Subtle differences leading to a different sql statements makes testing and not breaking things difficult.

I assume we know from the calling side what we want to call, so we might as well have different endpoints?

In the context for  fish-rfid, we will need something like "asve featureaction for samplingfeaturecode", but I was thinking of creating a dedicated endpoint that can call a dedicated function for this.

haakondr (approved):
sort of looks ok, but I am not convinced that we need a one-function-to-rule-them-all saving of featureactions

src/odm2_postgres_api/queries/core_queries.py
haakondr: this is indeed useful, just encountered the need for this (create featureaction for samplingfeaturecode).

haakondr, line 165: Tested this locally by rebasing this branch into my fish branch, and it crashes when I supply only samplingfeaturecode on line 165. Adapted this to the following and it works:

```
        return schemas.FeatureActions(samplingfeatureuuid=feature_action.samplingfeatureuuid,
                                      samplingfeaturecode=feature_action.samplingfeaturecode, **row)
```

haakondr: pushed a change to the branch with this fix

PR #64, opened 2020-09-11
Upgraded timescaledb to 17.4-pg12

haakondr:
major postgres version upgrade. Plan is to wipe our databases before
merging 49-metadata-setup branch so this is a good opportunity to
upgrade db as we have no regime for db updates without downtime yet

Question: we put our db files in /var/lib/postgresql/data/pg11. I
changed this to pg12 as that sort of makes sense. Is this is a best
practice and what is the use case? I assume that during breaking
upgrades we can manually move our data from pg<version> directory while
doing any possible migrations..?

question2: why is timescaledb update not picked up by renovatebot? :(

I skimmed through postgres 12 release notes and did not see any breaking changes that would affect us


haakondr:
also skimmed through timescaledb docs and do not find any breaking changes

PR #68, opened 2020-09-22
Reworked tests with rollback during test teardown

haakondr:
- created a new function scoped fixture "db_conn" that sets up a db
connection, starts a transaction and rollbacks during test teardown

- Refactored tests to not use dbpool and acquire connection but inject
it via the fixture

This speeds up tests by quite a lot as we no longer need to clear the db
tables.

TODO: I struggled with setting up init_metadata as a separate async
fixture as it relies on pytest-asyncio function scoped fixture
event_loop. I tried following the approach suggested at
https://github.com/pytest-dev/pytest-asyncio/issues/75 but did not
manage to "override" the fixture. Solved the issue by invoking
init_metadata within wait_for_db. Could perhaps rename this fixture, but
I think that ideally init_metadata is invoked as a separate fixture

TODO: In all our tests I have to explicitly import wait_for_db even
though db_conn is invoking it. Feels like I should not need to do this..

@RmStorm unrelated change/problem, but thought I'd also have a look at this now. When I invoke tests via pycharm the working directory is ./tests and then the pytest.ini addopts fail as it tries to find ./tests/tests/integration-test-docker-compose.yml. Is there a easy way to solve this within pytest-docker-compose?


tests/integration_test_fixtures.py
```diff
@@ -39,6 +44,7 @@ def wait_for_db(module_scoped_container_getter):
     try_db_init()
 
     asyncio.set_event_loop(asyncio.new_event_loop())
+    asyncio.run(init_metadata())
```
haakondr, line 47: I did not get the solution of redefining event_loop fixture to work so this is my workaround

tests/test_db.py
```diff
@@ -4,12 +4,12 @@
 import pytest
 
 from odm2_postgres_api.queries.user import create_or_get_user
-from integration_test_fixtures import wait_for_db, clear_db, init_dbpool
+from integration_test_fixtures import db_conn, wait_for_db
```
haakondr, line 7: I have to explicitly import wait_for_db even though db_conn fixture should invoke it. Am I doing something wrong?

tests/test_db.py
```diff
@@ -20,18 +20,16 @@ async def test_create_user(wait_for_db, clear_db):
         "updateTime": "2020-04-20T11: 45: 21.241Z",
         "roles": ["apps: admin", "niva"]
     }
-    db_pool = await init_dbpool()
-    async with db_pool.acquire() as connection:
-        # TODO: create obj and base64 encode in test instead of this, as this is not very readable
-        person = await create_or_get_user(connection, b64encode(json.dumps(user_obj).encode('utf-8')))
+    # TODO: create obj and base64 encode in test instead of this, as this is not very readable
+    person = await create_or_get_user(db_conn, b64encode(json.dumps(user_obj).encode('utf-8')))
```
haakondr, line 24: probably a good idea to ignore whitespace as I did a lot of re-indenting

tests/test_db.py
RmStorm: I'm not quite sure what is going on. integration_test_fixtures is not automatically imported by pytest I guess. Only the file `conftest.py` is automatically imported so that is where pytest expects fixtures.

What is so strange about it that running: `pytest --fixtures` does list the fixtures available in integration_test_fixtures...

tests/integration_test_fixtures.py
```diff
@@ -60,3 +66,16 @@ async def clear_db():
...
+async def db_conn(wait_for_db):
+    db_pool = await init_dbpool()
+    async with db_pool.acquire() as conn:
+        transaction = conn.transaction()
+
+        await transaction.start()
+        yield conn
+
+        # clean up after test
+        await transaction.rollback()
```
RmStorm, line 81: if an error happens in a test the rollback is not triggered, I think it should work like so:
```python
async with conn.transaction() as transaction:
    yield conn
    await transaction.rollback()
```
ugh except that it doesn't because the contextmanager doesn't return the transaction. I will make an issue in the asyncpg repo for it...

That means it has to be like so:
```python
transaction = conn.transaction()
await transaction.start()
try:
    yield conn
finally:
    await transaction.rollback()
```

tests/integration_test_fixtures.py
RmStorm: This works just fine.

RmStorm (changes requested):
I made some comments, currently the transactions is not always rolled back. If that is fixed it's good to go.

tests/integration_test_fixtures.py
haakondr: yeah that's how I originally had it, not sure why I changed it back without the try/finally, but we definitely need that, yes

RmStorm (approved):
This looks great. Let's merge.

PR #69, opened 2020-09-22
Added two new endpoints to find units and variables through api

RmStorm:
In order for both of these endpoints to be GET @zofiarudjord and I
decided not to use the UnitsCreate schema in the endpoint. The variable
endpoint naturally uses just a string in order to find the row.

We also did a small refactor to get rid of `get_unit` so that only
`find_unit` is left. This makes `find_unit` and `find_row` a little more
similar which seemed to be a little more readable.

### DIFF
--- a/src/odm2_postgres_api/routes/begroing_routes.py
+++ b/src/odm2_postgres_api/routes/begroing_routes.py
@@ -3,7 +3,7 @@ from collections import defaultdict
 
 from fastapi import Depends, Header, APIRouter
 
-from odm2_postgres_api.queries.core_queries import find_row, find_unit, get_unit
+from odm2_postgres_api.queries.core_queries import find_row, find_unit
 from odm2_postgres_api.routes.shared_routes import post_actions, post_results, post_categorical_results, \
     post_measurement_results
 from odm2_postgres_api.schemas.schemas import ProcessingLevels, UnitsCreate, Variables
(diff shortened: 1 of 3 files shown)


src/odm2_postgres_api/routes/begroing_routes.py
```diff
@@ -139,10 +139,12 @@ async def post_indices(new_index: schemas.BegroingIndicesCreate,
...
                                       "0", ProcessingLevels)
-    dimensionless_unit = await get_unit(connection, UnitsCreate(unitstypecv="Dimensionless", unitsabbreviation="-",
-                                                                unitsname="Dimensionless"))
-    seconds_unit = await get_unit(connection, UnitsCreate(unitstypecv="Time", unitsabbreviation="s",
-                                                          unitsname="second"))
+    dimensionless_unit = await find_unit(connection, UnitsCreate(unitstypecv="Dimensionless", unitsabbreviation="-",
+                                                                 unitsname="Dimensionless"), raise_if_none=True)
+    seconds_unit = await find_unit(connection, UnitsCreate(unitstypecv="Time", unitsabbreviation="s",
+                                                           unitsname="second"), raise_if_none=True)
+    if dimensionless_unit is None or seconds_unit is None:
```
RmStorm, line 146: This check has to be inserted to please the linter. The `raise_if_none=True` already ensures this but mypy doesn't understand. The other option is to comment line 158 & 174 with `  # type: ignore`.

haakondr (approved):

ZofiaRudjord (approved):
I tested the code with nivacloud-mass-spectromerty.

PR #71, opened 2020-09-23
Added a taxonomic classifier annotation table

RmStorm:
Also wrote a validation function to check validity of annotations.

It's not in use for all annotations yet so I think that should be done before merging this.


src/odm2_postgres_api/queries/core_queries.py
```diff
@@ -92,6 +92,18 @@ async def create_or_parse_annotations(conn: asyncpg.connection,
     return annotation_ids
 
 
+async def insert_taxonomic_classifier(conn: asyncpg.connection, taxon: schemas.TaxonomicClassifierCreate):
+    taxon_data = {k: v for k, v in taxon if k != "annotations"}
+    async with conn.transaction():
+        taxon_row = await conn.fetchrow(make_sql_query('taxonomicclassifiers', taxon_data), *taxon_data.values())
+        logging.info(taxon_row)
+        for annotation_id in await create_or_parse_annotations(conn, taxon.annotations):
```
haakondr, line 100: I am not very found of functions like create_or_parse_annotations that take multiple types of input. In this case it will always do a sql query (?), but in other usages it will take annotation_ids. Could we make two different functions?

Could consider to create a slightly different abstraction here:

```
save_annotations(conn, annotations: List[AnnotationsCreate], referenced_table: str, referenced_id: int) -> List[Annotations]

# usage:
save_annotations(conn, taxon.annotations, 'TaxonomicClassifiersAnnotations', taxon.taxonomicclassifierid)
```

I imagine the save_annotations function would make sure to insert annotations and the bridge tables

src/odm2_postgres_api/schemas/schemas.py
```diff
@@ -269,6 +276,10 @@ class MethodsCreate(BaseModel):
     # annotation_id_list: List[int] = []  # This requires some thought, are both this field and annotation allowed?
     annotations: List[Union[AnnotationsCreate, int]] = []
 
+    @validator('annotations')
+    def check_annotations(cls, annotations):
+        return create_annotation_type_cv_checker(annotations, "Method annotation")
```
haakondr, line 281: if we always know what the annotation cv type should be, does it make sense to have the client provide it? With this validation we always require it to be "Method annotation"

haakondr (approved):
Looks ok. Made some comments, I think there are some nitpicks that should be resolved or discussed before merging

haakondr:
It should be quite straight forward to start writing tests now, so I think changes should have integration tests setup

src/odm2_postgres_api/metadata_init/data/general/controlled_vocabularies.py
```diff
@@ -10,6 +10,12 @@
                   "and measured over time.",
     "category": "Coverage",
     "controlled_vocabulary_table_name": "cv_resulttype"
+}, {
+    "term": "taxonomicClassifierAnnotation",
+    "name": "Taxonomic classifier annotation",
+    "definition": "An annotation or qualifying comment about an TaxonomicClassifier.",
+    "category": "Annotation",
+    "controlled_vocabulary_table_name": "cv_annotationtype"
```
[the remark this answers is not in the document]
RmStorm: yeah I misunderstood this, sometimes sampling features or annotated with speciesannotationtype since a samplingfeature can be different things. I've removed some of the checks on it that I wrote.

src/odm2_postgres_api/db_init/ODM2_for_PostgreSQL.sql
```diff
@@ -2330,4 +2330,10 @@ ALTER TABLE ODM2.personexternalidentifiers
...
 ALTER TABLE ODM2.externalidentifiersystems ADD UNIQUE (externalidentifiersystemname);
-ALTER TABLE ODM2.units ADD UNIQUE (unitstypecv, unitsabbreviation);
\ No newline at end of file
+ALTER TABLE ODM2.units ADD UNIQUE (unitstypecv, unitsabbreviation);
+
+create table ODM2.TaxonomicClassifiersAnnotations (
+	bridgeid serial  NOT NULL primary key,
+	taxonomicclassifierid integer  NOT NULL,
+	annotationid integer  NOT NULL
+);
```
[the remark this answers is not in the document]
RmStorm: thanks, good catch.

src/odm2_postgres_api/queries/core_queries.py
RmStorm: I sortoff agree but it's used in a bunch of places now. It works quite well but if you think it's to unreliable we should make an issue and resolve it in a different PR.

Also overloaded functions are not that weird a concept right?

src/odm2_postgres_api/schemas/schemas.py
RmStorm: see previous comment. I misunderstood how this can be used and this check is in error. I have removed it.

PR #74, opened 2020-09-28
Writing new observations to aquamonitor API

haakondr:
I think this is in a semi-workable state right now. I just code reviewed myself now, and made some notes about missing parts.

NB: the main function in the aquamonitor_client.py does currently write, then delete (not in a transaction). This is not 100% safe, but I was double checking the database afterwards and made sure to not leave anything behind. This is not ideal, but hopefully enables us to move forward without re-inventing too many wheels.

As it is now, I have introduced a simple environment variable WRITE_TO_AQUAMONITOR which will typically be false for all environments except production. 

## changes
- Created API wrapper for aquamonitor API (aquamonitor client) based on
httpx AsyncClient

- manually constructed the API request/response types from aquamonitor
API in aquamonitor_api_types.py. Kept the same names from the API with
the exception of some fields that use leading underscore which did not
work very well with pydantic. Mapped these using aliases, but the
serialized json remains the same
- hardcoded a map between method name and nivabase id. There is no
endpoint to fetch method based on method name currently, so doing this
for now. May consider to talk with Roar to add a new endpoint in the API
for this
- Constructed a new data structure for begroing observations where we
have a list of (taxon, method, value). This is easier to work with and I
think this should be moved over to the frontend
- Roar has created a system user to interact with aquamonitor API. I
stored this as k8s secret aquamonitor-api-user

## todos
- I am calling a DELETE after each write. This is not 100% safe, but
also very pragmatic. May consider to opt for a safer regime when testing
- I made some helper utils for logging outgoing requests, adding
span-id/trace-id etc that should go to nivacloud-logging
- station codes are not 1:1 with nivabase station codes. I think we may
need to query the BEGROING.IMPORT_STATIONS table to get the actual
station code? this needs to be discussed
- need to actually integrate this into the endpoint. I am just calling
this manually now
- not sure what to do in such situations as we'd have two
PROJECTS_STATIONS rows in nivabase. Don't want to duplicate values so I
just throw an error for now

Related issue:
https://github.com/NIVANorge/begroing-data-upload/issues/60


src/odm2_postgres_api/aquamonitor/aquamonitor_client.py
```diff
@@ -0,0 +1,244 @@
...
+    BegroingSampleCargoCreate, StationCargo, BegroingObservationCargoCreate, MethodCargo, TaxonomyCodeCargo, \
+    ProjectCargo, ProjectCargoCreate
+from odm2_postgres_api.aquamonitor.aquamonitor_mapping import METHODS_NIVABASE_MAP
+from odm2_postgres_api.schemas.schemas import BegroingResultCreate, Directive, SamplingFeatures, Methods, \
+    BegroingObservations, BegroingObservationValues, TaxonomicClassifier
+
+
+async def get_taxonomy_codes(client: AsyncClient, domain_id: str):
+    response = await client.get(f"/taxonomy/domains/{domain_id}/codes")
+    response.raise_for_status()
```
haakondr, line 22: I may consider to introduce a httpx event_hook that calls  raise_for_status on response. event_hook is a brand new concept in httpx 0.15 that did not exist when I started out

src/odm2_postgres_api/aquamonitor/aquamonitor_client.py
```diff
@@ -0,0 +1,244 @@
...
+
+from odm2_postgres_api.aquamonitor.aquamonitor_api_types import BegroingSampleCargo, BegroingObservationCargo, \
+    BegroingSampleCargoCreate, StationCargo, BegroingObservationCargoCreate, MethodCargo, TaxonomyCodeCargo, \
+    ProjectCargo, ProjectCargoCreate
+from odm2_postgres_api.aquamonitor.aquamonitor_mapping import METHODS_NIVABASE_MAP
+from odm2_postgres_api.schemas.schemas import BegroingResultCreate, Directive, SamplingFeatures, Methods, \
+    BegroingObservations, BegroingObservationValues, TaxonomicClassifier
+
+
+async def get_taxonomy_codes(client: AsyncClient, domain_id: str):
```
haakondr, line 20: Missing type defs. Will add these in

src/odm2_postgres_api/aquamonitor/aquamonitor_client.py
```diff
@@ -0,0 +1,244 @@
...
+async def get_begroing_observations(client: AsyncClient, sample_id: int, method_id="null") \
+        -> List[BegroingObservationCargo]:
+    response = await client.get(f"/begroing/samples/{sample_id}/observations?methodId={method_id}")
+    response.raise_for_status()
+    return [BegroingObservationCargo(**o) for o in response.json()]
+
+
+async def post_begroing_sample(client: AsyncClient, sample: BegroingSampleCargoCreate) -> BegroingSampleCargo:
+    # TODO: this does not  set json headers properly..
+    response = await client.post(f"/begroing/samples", data=sample.json(), headers={"Content-Type": "application/json"})
```
haakondr, line 57: will look at how to properly post json with pydantic objects. Since pydantic classes ship with .json function I don't think this is called when using json=

src/odm2_postgres_api/aquamonitor/aquamonitor_client.py
```diff
@@ -0,0 +1,244 @@
...
+
+async def delete_begroing_sample(client: AsyncClient, begroing_sample_id: int):
+    response = await client.delete(f"/begroing/samples/{begroing_sample_id}")
+    response.raise_for_status()
+
+
+async def get_project_stations(client: AsyncClient, project_name: str, station_code: str) -> Optional[StationCargo]:
+    res = await client.get(f"/query/Stations?projectName={project_name}&stationCode={station_code}")
+    if res.status_code == 404:
+        raise HTTPException(422, f"Did not find station={station_code} in Aquamonitor")
```
haakondr, line 70: hmm. not sure if it makes sense to re-map this to status 422. Maybe keep it as 404

src/odm2_postgres_api/aquamonitor/aquamonitor_client.py
```diff
@@ -0,0 +1,244 @@
...
+
+async def get_project(client: AsyncClient, project: Directive) -> ProjectCargo:
+    res = await client.get(f"/query/projects")
+    res.raise_for_status()
+    projects = [ProjectCargo(**p) for p in res.json()]
+
+    actual_project = [p for p in projects if p.name == project.directivedescription]
+    # TODO: this is very ad hoc. Need to validate that we find a project
+    # TODO: should we store reference to project id in odm2 to query more efficiently ?
+    return actual_project[0]
```
haakondr, line 83: TODO: should validate that len(actual_project) == 1

src/odm2_postgres_api/aquamonitor/aquamonitor_client.py
```diff
@@ -0,0 +1,244 @@
...
+
+
+async def get_method(client: AsyncClient, method_name: str, unit: str, laboratory: str):
+    res = await client.get(f"/query/Methods?name={method_name}&unit={unit}&laboratory={laboratory}&methodRef=")
+    res.raise_for_status()
+    return res.json()
```
haakondr, line 89: unused method, but has an API endpoint. I guess it makes sense to leave it for the future?

src/odm2_postgres_api/aquamonitor/aquamonitor_client.py
```diff
@@ -0,0 +1,244 @@
...
+
+
+async def update_begroing_observation(client: AsyncClient,
+                                      observation: BegroingObservationCargo) -> BegroingObservationCargo:
+    sample_id = observation.Sample.Id
+    res = await client.put(f"/begroing/samples/{sample_id}/observations/{observation.Id}",
+                           data=observation.json(),
+                           headers={"Content-Type": "application/json"})
+    res.raise_for_status()
+    return res.json()
```
haakondr, line 129: TODO: Have not tested this yet

src/odm2_postgres_api/aquamonitor/aquamonitor_client.py
```diff
@@ -0,0 +1,244 @@
...
+
+
+async def post_begroing_observations(client: AsyncClient, result: BegroingObservations):
+    station_code = result.station.samplingfeaturecode
+    station: StationCargo = await get_project_stations(client, result.project.directivedescription, station_code)
+
+    sample = await get_or_create_begroing_sample(client, station, result)
+
+    created_observersions = await asyncio.gather(
+        *[post_begroing_observation(client, sample, o) for o in result.observations])
```
haakondr, line 163: doing writes in async. Not sure how robust aquamonitor/nivabasen is, but I suppose it can handle this?

src/odm2_postgres_api/aquamonitor/aquamonitor_client.py
```diff
@@ -0,0 +1,244 @@
...
+    observation_ids = [o.Id for o in created_observersions]
+    logging.info("Successfully stored observations", extra={"observation_ids": observation_ids})
```
haakondr, line 165: TODO: missing return statement

src/odm2_postgres_api/aquamonitor/aquamonitor_client.py
```diff
@@ -0,0 +1,244 @@
...
+                                                                    observations=observation_values))
+
+
+if __name__ == '__main__':
+    username = os.environ["AQUAMONITOR_USER"]
+    password = os.environ["AQUAMONITOR_PASSWORD"]
+    os.environ["HTTPX_LOG_LEVEL"] = "debug"
+    setup_logging(plaintext=True)
+    asyncio.run(main(api_url="https://test-aquamonitor.niva.no/AquaServices/api",
+                     username=username, password=password))
```
haakondr, line 244: this needs to be cleaned up and probably deleted

src/odm2_postgres_api/schemas/schemas.py
```diff
@@ -541,12 +541,26 @@ class MeasurementResults(MeasurementResultsCreate):
...
+    taxon: TaxonomicClassifier
+    method: Methods
+    value: str
+
+
+class BegroingObservations(BaseModel):
+    project: Directive
+    date: dt.datetime
+    station: SamplingFeatures
+    observations: List[BegroingObservationValues]
```
haakondr, line 561: this is my proposed new structure. I think this makes sense also in the frontend, so that we can do the mapping there

src/odm2_postgres_api/schemas/schemas.py
```diff
@@ -541,12 +541,26 @@ class MeasurementResults(MeasurementResultsCreate):
...
     date: dt.datetime
-    station: Dict
+    station: SamplingFeatures
     taxons: List[Dict]
     methods: List[Methods]
     observations: List[List[str]]
 
 
+class BegroingObservationValues(BaseModel):
+    # TODO: ADD cv to graphql query
```
haakondr, line 551: (graphql query is missing taxonomicclassifiertypecv so the frontend does not have this)

haakondr:
I have not added any tests, but this surely could use some test coverage. Not sure what the best approach would be, but could maybe be a good fit with either stubbing or a http mock ala http://wiremock.org/ to stub out aquamonitor API

src/odm2_postgres_api/aquamonitor/aquamonitor_client.py
danielborseth: Is there a way to get stations using the project id?

danielborseth: I don't think its necessary to use HTTP status code semantics when you're the one consuming the API, or if you own both sides of the contract. A _404 this resource does not exist_ should be enough.

danielborseth:
I've just created aquamonitor as an external service on the test cluster. It should be reachable using _aquamonitor-api_

src/odm2_postgres_api/aquamonitor/aquamonitor_client.py
haakondr: yes I suppose you are right. Not sure why I landed on this

haakondr:
> _aquamonitor_

Ok, great. Should be added to source code so that it is applied automatically. 

@RmStorm is it as simple as adding a file in k8s/services/aquamonitor-api.yaml ?

src/odm2_postgres_api/aquamonitor/aquamonitor_client.py
haakondr: It is possible to retrieve stations using project id, but we have not currently stored nivabase project id's in odm2.

Question is, should we store nivabase id's in the new database?

## pros

- easy to identify the exact project
- robust. Easier to deal with changes to identifiers (what happens if project name changes?)

## cons

- we spread a lot of nivabase internals into the new database

I feel that referring to nivabase data with unique identifiers that are not sequential identifiers is more robust, but it could be that not all data is uniquely identifiable without their primary key

danielborseth:
Here's the k8s manifest https://github.com/NIVANorge/nivacloud-manifests/tree/master/aquamonitor

src/odm2_postgres_api/aquamonitor/aquamonitor_client.py
danielborseth: Looks like project names should be unique in nivadatabase, but surrogate keys are always easier to deal with. What happens here when someone creates a project with the same name? I would find that weird.

RmStorm (approved):
I have made a bunch of minor comments that you can choose to ignore or implement.
Do you want to factor out the new logging&tracing utilities to the logging library? Or do you want to merge first? Same goes for the refactored structure of the begroing observations. Merge first or refactor before merging?

src/odm2_postgres_api/aquamonitor/aquamonitor_client.py
haakondr: no reason. I think I did a search/replace for "res.raise_for_status()", but in this case it was response.raise_for_status(). Just a mistake

src/odm2_postgres_api/aquamonitor/aquamonitor_client.py
```diff
@@ -0,0 +1,231 @@
...
+
+async def get_taxonomy(client: AsyncClient, domain_name: str, code: str) -> TaxonomyCodeCargo:
+    res = await client.get(f"/query/taxonomy/domains?domain={domain_name}&code={code}")
+    handle_aquamonitor_error(res)
+    return TaxonomyCodeCargo(**res.json())
+
+
+async def get_taxonomy_domain_id(client: AsyncClient, domain: str):
+    response = await client.get(f"/query/taxonomy/domains?domain={domain}")
+    response.raise_for_status()
```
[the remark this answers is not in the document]
haakondr: just a mistake from my end

src/odm2_postgres_api/aquamonitor/aquamonitor_client.py
```diff
@@ -0,0 +1,231 @@
...
+        created_observations = await asyncio.gather(
+            *[post_begroing_observation(client, sample, o) for o in result.observations])
+        observation_ids = [o.Id for o in created_observations]
+
+        logging.info("Successfully stored observations", extra={"observation_ids": observation_ids})
+        return {"sample": sample, "observations": created_observations}
+
+
+# TODO: move this to nivacloud-logging
+async def traced_request(request):
```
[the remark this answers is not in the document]
haakondr: I prefer to do it after PR

src/odm2_postgres_api/aquamonitor/aquamonitor_api_types.py
```diff
@@ -0,0 +1,113 @@
...
+    Matrix: Optional[str] = None
+
+
+class DomainCargo(BaseModel):
+    Id: int
+    Name: Optional[str] = None
+    Description: Optional[str] = None
+
+
+class TaxonomicLevelCargo(BaseModel):
```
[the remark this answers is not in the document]
haakondr: Tested this and we need the updateforwardref stuff.

```
from odm2_postgres_api.aquamonitor.aquamonitor_api_types import TaxonomicLevelCargo

parent = TaxonomicLevelCargo(Id=2, Name="parent")
child = TaxonomicLevelCargo(Id=1, Name="child", Parent=parent)
print(child)
```
throws

pydantic.errors.ConfigError: field "Parent" not yet prepared so type is still a ForwardRef, you might need to call TaxonomicLevelCargo.update_forward_refs().

PR #78, opened 2020-10-07
reformatted all source code with black

haakondr:
I ran "black ." which formatted all the files with black.

added a pyproject.toml. This one only contains black config for now, but
I propose that we move towards using pyproject.toml + poetry for
environment management.

- set line-length to 120
- removed pycodestyle
  - black does not seem to be 100% strict on line length so sometimes it
overflows the max length limit. I think this is ok, but does not fly
well with running pycodestyle in addition. I suggest we just run black
--check and see how that goes. Ref comment
psf/black#1314 (comment)

Recommended setup:

- everyone setup file watchers in pycharm
- we run "black --check ." as part of docker build (dockerfile)

this is my setup in pycharm:
file-watcher-black


RmStorm (changes requested):
This is really great, we've needed an autoformatter for a long while now. Black is very opinionated though. Can you spend a little time on investigating autopep8 and yapf as well? I think yapf is much more configurable and could suite our needs much better!

RmStorm (approved):
Looks great. We should implement this in the other repo's as well and get everyone to configure Pycharm correctly.

PR #79, opened 2020-10-07
yapf formatting

haakondr:
ran with the following command:

yapf --recursive -i src/


RmStorm (changes requested):
I think I prefer this one, it formats comprehensions slightly better and it offers us the ability to customize it. We just need to remove the black command from the Dockerfile before merging.

PR #81, opened 2020-10-07
Add a RelatedTaxonomicClassifiers table

### LINKED ISSUE #80: Add ability to create synonyms between taxonomicclassifiers
After discussing with @jeved and people from begroing it has become quite clear that we need slightly more complicated taxon management than supplied by ODM2's single table implementation. As a first step I will implement a `related_taxonomic_classifiers` table that is exactly analogous to e.g. `related_actions` as can be seen in [this schema](http://odm2.github.io/ODM2/schemas/ODM2_Current/diagrams/ODM2Core.html).

This is important for actually moving the begroing data over to ODM2 since otherwise we would have no way to retain the synonyms. Retaining synonyms is important because synonyms can be wrongly assigned!

In the first version of the [begroing transfer script](https://github.com/NIVANorge/begroing-data-upload/blob/dev/begroing_python_utils/transfer_stations.py) I simply resolved the synonym links and left the species that were no longer 'accepted' out of ODM2. However as @jeved pointed out, sometimes a synonym is assigned and then later it is assigned differently or the decision is reverted. For this reason we must keep track of the original classifications for any observation otherwise such a reversal can never be accommodated by us.

There is a number of other issues in taxon management but they should get their own issues and we should decide on a feature by feature basis if we need to implement them now or if it can wait.

RmStorm:
This table links taxonomicclassifiers using the standard relationship
table. There are several precendents in the ODM2 schema such as the
relatedactions table as can be seen here:
http://odm2.github.io/ODM2/schemas/ODM2_Current/diagrams/ODM2Core.html

issue #80 describes why we need this table, closes #80


tests/test_core_queries.py
```diff
@@ -17,43 +18,47 @@
...
+@pytest.mark.docker
+@pytest.mark.asyncio
+async def test_create_taxons_with_annotations(db_conn):
+    for row in BEGROING_TAXON_ROWS[:-1]:
+        data = format_taxon_data(row)
+        taxon_result = await insert_taxonomic_classifier(db_conn, schemas.TaxonomicClassifierCreate(**data))
+        assert type(taxon_result) == schemas.TaxonomicClassifier
+
+    with pytest.raises(ForeignKeyViolationError):  # 0 can never exist since 'serial' datatype starts at 1
+        data = format_taxon_data(BEGROING_TAXON_ROWS[-1], [(0, 'Is previous version of')])
```
haakondr, line 51: I suggest we construct test data as part of the actual test.

What happens if one test mutates BEGROING_TAXON_ROWS? That would break others tests wouldn't it?

haakondr: I suggest we construct a helper function that constructs a taxon with default values that can be overridden. Something like:

```
def create_taxon(taxon_data):
    t = {
        "Artsliste_Begroing_Sys_Id": 1,
        "rubin_kode": "AC AU;HE",
        "latinsk_navn": "Achnanthes austriaca var. helvetica",
        "autor": "Hust.",
        "autor_ref": None,
        "synonym": "ACHN HEL",
        "takson_type": "VARIET",
        "eier_takson": "BACILLAR",
        "ph_opt": 5.4,
        "ph_ref": "Stevenson",
        "Kommentar": None,
        "Reg_Dato": None,
        "Reg_Av": None,
        "Mod_Dato": "2016-02-04 15:56:28",
        "Mod_Av": None,
        "FF-verdi": None,
        "rubin_nr": None,
        "merke": None,
        "Field1": "S",
        "egenskaper": "5,4",
        "dato": "2002-12-09 00:00:00",
        "RegHvor": "Stevenson",
        "Typesamling": None,
        "Foto/tegning": None,
        "ArtslisteId": 1,
    }

    return {**t, **taxon_data}
```

and even better if we make a pydantic class (perhaps not needed for these annotation types)

tests/test_core_queries.py
```diff
@@ -17,43 +18,47 @@
...
+        await insert_taxonomic_classifier(db_conn, schemas.TaxonomicClassifierCreate(**data))
+
+    # Testing that an insert with a valid related action succeeds
+    data = format_taxon_data(BEGROING_TAXON_ROWS[-1], [(taxon_result.taxonomicclassifierid, 'Is previous version of')])
+    await insert_taxonomic_classifier(db_conn, schemas.TaxonomicClassifierCreate(**data))
```
haakondr, line 56: could also be expanded to test the following:

1. returned object from insert_taxonomic_classifier
2. query afterwards to check if the data was really written

haakondr (approved):

haakondr:
I have not tested this, but maybe we can do that together?

PR #92, opened 2020-10-26
Fetch begroing data

haakondr:
## changes 

- new endpoint GET /station/{sampling_feature}/project/{project_id}/start/{start_time}/end/{end_time}
- fixed POST begroing results so that results.date is also stored

could need some more polish, but I would like to start on the frontend.

### DIFF
--- a/src/odm2_postgres_api/queries/begroing_queries.py
+++ b/src/odm2_postgres_api/queries/begroing_queries.py
@@ -1,0 +1,68 @@
+import uuid
+from datetime import datetime
+from typing import Dict
+
+from asyncpg import connection
+
+from odm2_postgres_api.queries.core_queries import find_row
+from odm2_postgres_api.schemas.schemas import (
+    BegroingObservations,
+    Directive,
+    SamplingFeatures,
+    BegroingObservationValues,
+    TaxonomicClassifier,
+    Methods,
+)
+
+
+async def find_begroing_results(
+    conn: connection, project_id: int, sampling_feature: uuid, date: datetime
+) -> BegroingObservations:
+    def observations_from_row(row: Dict) -> BegroingObservationValues:
+        taxon = TaxonomicClassifier(**row)
+        method = Methods(**row)
+        value = row["categorical_value"] or str(row["measurement_value"])
+        return BegroingObservationValues(taxon=taxon, method=method, value=value)
+
+    project: Directive = await find_row(conn, "directives", "directiveid", project_id, Directive, raise_if_none=True)
+    sampling_feature: SamplingFeatures = await find_row(
+        conn, "samplingfeatures", "samplingfeatureuuid", sampling_feature, SamplingFeatures, raise_if_none=True
+    )
+
+    rows = await conn.fetch(
+        """
+select crv.datavalue as categorical_value,
+       mrv.datavalue as measurement_value,
+       t.*,
+       m.*,
+       aff.primaryemail
+from results r
+         full outer join categoricalresultvalues crv on crv.resultid = r.resultid
+         full outer join measurementresultvalues mrv on mrv.resultid = r.resultid
+         inner join taxonomicclassifiers t on t.taxonomicclassifierid = r.taxonomicclassifierid
+         inner join featureactions fa on r.featureactionid = fa.featureactionid
+         inner join actions a on fa.actionid = a.actionid
+         inner join methods m on a.methodid = m.methodid
+         inner join actiondirectives ad on ad.actionid = a.actionid
+         inner join actionby ab on ab.actionid = a.actionid
+         inner join affiliations aff on ab.affiliationid = aff.affiliationid
+         inner join samplingfeatures s on fa.samplingfeatureid = s.samplingfeatureid
+where 
+    fa.samplingfeatureid = $1 and
+    ad.directiveid = $2 and 
+    r.resultdatetime=$3 and r.resultdatetimeutcoffset=$4
+    """,
+        sampling_feature.samplingfeatureid,
+        project_id,
+        date,
+        date.utcoffset() or 0,
+    )
+
+    observations = [observations_from_row(r) for r in rows]
+
+    return BegroingObservations(
+        project=project,
+        date=date,
+        station=sampling_feature,
+        observations=observations,
+    )
--- a/src/odm2_postgres_api/queries/begroing_queries.py
+++ b/src/odm2_postgres_api/queries/begroing_queries.py
@@ -16,6 +16,19 @@ from odm2_postgres_api.schemas.schemas import (
 )
 
 
+def observations_from_row(station: SamplingFeatures, project: Directive, row: Dict) -> BegroingObservation:
+    return BegroingObservation(
+        project=project,
+        station=station,
+        timestamp=row["resultdatetime"],
+        taxon=TaxonomicClassifier(**row),
+        method=Methods(**row),
+        categorical_value=row["categorical_value"],
+        measurement_value=row["measurement_value"],
+        flag=row["censorcodecv"],
+    )
+
+
 async def find_begroing_results(
     conn: connection,
     project_id: int,
@@ -23,13 +36,6 @@ async def find_begroing_results(
     start_time: datetime,
     end_time: datetime,
 ) -> List[BegroingObservation]:
-    def observations_from_row(station: SamplingFeatures, project: Directive, row: Dict) -> BegroingObservation:
-        taxon = TaxonomicClassifier(**row)
-        method = Methods(**row)
-        value = row["categorical_value"] or str(row["measurement_value"])
-        return BegroingObservation(
-            project=project, station=station, timestamp=row["resultdatetime"], taxon=taxon, method=method, value=value
-        )
 
     project: Directive = await find_row(conn, "directives", "directiveid", project_id, Directive, raise_if_none=True)
     sampling_feature: SamplingFeatures = await find_row(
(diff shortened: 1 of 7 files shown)


src/odm2_postgres_api/routes/begroing_routes.py
```diff
@@ -123,13 +145,14 @@ async def post_begroing_result(
...
             for result_index in method_observations:
-
                 data_result = schemas.ResultsCreate(
                     samplingfeatureuuid=begroing_result.station.samplingfeatureuuid,
                     actionid=completed_action.actionid,
                     resultuuid=str(uuid.uuid4()),
                     resulttypecv=result_type_and_unit_dict[method.methodname][0],
                     variableid=abundance_variable.variableid,
+                    resultdatetime=begroing_result.date,
+                    resultdatetimeutcoffset=0,
```
haakondr, line 155: utcoffset is already validated to match 0, right @RmStorm ?

haakondr: added this so that time is also stored on results

src/odm2_postgres_api/schemas/schemas.py
```diff
@@ -589,6 +589,15 @@ class BegroingObservations(BaseModel):
     observations: List[BegroingObservationValues]
 
 
+class BegroingObservation(BaseModel):
+    project: Directive
+    timestamp: dt.datetime
+    station: SamplingFeatures
+    taxon: TaxonomicClassifier
+    method: Methods
+    value: str
```
haakondr, line 598: this is somewhat duplicate of the other types. We already have BegroingObservations type which is used by the write endpoint when posting to aquamonitor API. I plan on refactoring this so that we use the same type.

This one is fully denormalized and I think that is the most flexible data model

tests/test_begroing.py
```diff
@@ -52,49 +51,55 @@ async def test_post_new_indices(db_conn):
...
+
+    return TaxonomicClassifierCreate(**t)
+
+
 @patch("odm2_postgres_api.routes.begroing_routes.store_begroing_results", autospec=True)
 @pytest.mark.asyncio
 @pytest.mark.docker
 async def test_post_new_begroing_observations(store_begroing_results, db_conn):
-    taxon_create = TaxonomicClassifierCreate(
+    taxa_create = [generate_taxon() for i in range(0, 3)]
```
haakondr, line 70: I was planning on writing some helper functions for posting data so that we can split the tests for the GET endpoint into separate tests, but have not done that yet

src/odm2_postgres_api/queries/begroing_queries.py
```diff
@@ -0,0 +1,70 @@
...
+         inner join methods m on a.methodid = m.methodid
+         inner join actiondirectives ad on ad.actionid = a.actionid
+         inner join actionby ab on ab.actionid = a.actionid
+         inner join affiliations aff on ab.affiliationid = aff.affiliationid
+         inner join samplingfeatures s on fa.samplingfeatureid = s.samplingfeatureid
+where
+    fa.samplingfeatureid = $1 and
+    ad.directiveid = $2 and
+    r.resultdatetime>=$3 and
+    r.resultdatetime<=$4
```
haakondr, line 62: I am not checking that utcoffset is matching. Not sure how to best do this. I think the whole concept of having utcoffsets as separate columns is very cumbersome. 

Yes, I get that odm2 wants to be db agnostic

RmStorm: I agree, I'm not too sure what to do about it though.

src/odm2_postgres_api/queries/begroing_queries.py
```diff
@@ -0,0 +1,70 @@
...
+    TaxonomicClassifier,
+    Methods,
+    BegroingObservation,
+)
+
+
+async def find_begroing_results(
+    conn: connection,
+    project_id: int,
+    sampling_feature: uuid,
```
RmStorm, line 22: This variable should be called something else, maybe `sampling_feature_uuid`. Otherwise mypy will complain about redeclaration on line 35. Also `uuid` is not a valid type. Should be `uuid.UUID`.

src/odm2_postgres_api/queries/begroing_queries.py
```diff
@@ -0,0 +1,70 @@
...
+    start_time: datetime,
+    end_time: datetime,
+) -> List[BegroingObservation]:
+    def observations_from_row(station: SamplingFeatures, project: Directive, row: Dict) -> BegroingObservation:
```
RmStorm, line 26: I don't think this function should be nested. Also (just a preference), I like this a little better :see_no_evil: :
```python
def observations_from_row(station: SamplingFeatures, project: Directive, row: Dict) -> BegroingObservation:
    return BegroingObservation(
        project=project,
        station=station,
        timestamp=row["resultdatetime"],
        taxon=TaxonomicClassifier(**row),
        method=Methods(**row),
        value=row["categorical_value"] or str(row["measurement_value"])
    )
```

src/odm2_postgres_api/queries/begroing_queries.py
```diff
@@ -0,0 +1,70 @@
...
+        """
+select crv.datavalue as categorical_value,
+       mrv.datavalue as measurement_value,
+       r.resultdatetime,
+       t.*,
+       m.*,
+       aff.primaryemail
+from results r
+         full outer join categoricalresultvalues crv on crv.resultid = r.resultid
+         full outer join measurementresultvalues mrv on mrv.resultid = r.resultid
```
RmStorm, line 49: I think these should be left joins? instead of full outer joins?

src/odm2_postgres_api/routes/begroing_routes.py
```diff
@@ -48,6 +52,24 @@
...
+)
+async def get_begroing_results(
+    sampling_feature_uuid: uuid.UUID,
+    project_id: int,
+    start_time: datetime,
+    end_time: datetime,
+    connection=Depends(api_pool_manager.get_conn),
+    niva_user=Header(None),
+) -> List[BegroingObservation]:
+    user = await create_or_get_user(connection, niva_user)
```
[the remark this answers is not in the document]
RmStorm: middleware is a nice idea! I also think we can be quite verbose with logging. We can always later change statements to debug or remove them when we have more traffic.

src/odm2_postgres_api/routes/begroing_routes.py
RmStorm: yes, looks good.

tests/test_begroing.py
```diff
@@ -52,49 +51,55 @@ async def test_post_new_indices(db_conn):
...
+            "methodcode": "begroing_4",
+            "methoddescription": "A quantitative observation is made assessing the abundance of a species in...",
+            "methodid": 6,
+            "methodlink": None,
+            "methodname": "Macroscopic coverage",
+            "methodtypecv": "Observation",
+        }
+    )
+    microscoping_abundance = Methods(
+        **{
```
RmStorm, line 84: the intermediate dict seems a little unnecessary? is there some gotcha with using keywords directly on the pydantic object?

RmStorm (changes requested):
Looks great! Currently there is a linter issue that has to be fixed (see first comment). In addition to that I made commented with some minor nitpickings. These other comments you can ignore if you prefer it the way it currently is.

src/odm2_postgres_api/queries/begroing_queries.py
haakondr: yes you are right

tests/test_begroing.py
haakondr: hm. I think I thought it was easier to read by defining it as dicts

haakondr:
@RmStorm I have made some changes, added a csv endpoint and a function that constructs a csv file from the list of observations.

This mapping is not 100%, but I would like to get some feedback at this stage. For instance, I am setting flag to "Less than" instead of ">"

tests/testdata_builders.py
```diff
@@ -53,7 +53,7 @@ def generate_taxon_create(overrides: Dict = {}) -> TaxonomicClassifierCreate:
     return TaxonomicClassifierCreate(**{**t, **overrides})
 
 
-def default_method(overrides: Dict) -> Methods:
+def default_method(overrides: Dict = {}) -> Methods:
```
RmStorm, line 56: It is very bad practice to use a mutable value as default argument! It is only declared once at the function declaration not once per function invocation. This means that all function calls have access to the same mutable object!

src/odm2_postgres_api/schemas/schemas.py
```diff
@@ -595,12 +595,27 @@ class BegroingObservation(BaseModel):
...
+        if mv is not None and cv is not None:
+            raise ValueError("Both measurement_value and categorical_value cannot be set at the same time")
+
+        if mv is None and cv is None:
+            raise ValueError("Either measurement_value or categorical_value needs to be defined")
+
+        return values
+
+    @root_validator
+    def validate_flag(cls, values):
```
RmStorm, line 615: good catch!

RmStorm (changes requested):
I know it's just in a test but using mutables as default values is asking for trouble! Otherwise this looks great.

tests/test_begroing.py
RmStorm: Completely fine by me!

tests/testdata_builders.py
haakondr: hm. you are right.

How would I do this then? I'd like to declare a default value as an empty object

haakondr: should I set default=None and then do this?

```
o = overrides or {}
Methods(**{**default, **o})
```

RmStorm: yes, defaulting to None and using a ternary is typical.
``` python
o = {} if overrides is None else overrides
```
This is a little more certain in case overrides is supplied but falsy.

haakondr:
Closing PR. While I put quite a lot of work into this, I don't think this is the way

PR #96, opened 2020-10-29
Update to tabluar format

roryscot:
Change backend to fit with new format: https://github.com/NIVANorge/begroing-data-upload/issues/148

### DIFF
--- a/src/odm2_postgres_api/routes/begroing_routes.py
+++ b/src/odm2_postgres_api/routes/begroing_routes.py
@@ -198,7 +198,7 @@ async def post_begroing_result(
 
 @router.post("/indices", response_model=schemas.BegroingIndices)
 async def post_indices(
-    new_indices: schemas.BegroingIndicesCreate,
+    data: schemas.BegroingIndicesCreate,
     connection=Depends(api_pool_manager.get_conn),
     niva_user: str = Header(None),
 ) -> BegroingIndices:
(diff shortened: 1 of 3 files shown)


src/odm2_postgres_api/schemas/schemas.py
```diff
@@ -593,16 +594,22 @@ class BegroingResult(BegroingResultCreate):
     personid: int
 
 
-class IndicesInfo(BaseModel):
-    indexType: str
-    indexValue: float
+class BasicIndexingUnit(Dict):
+    EQR: Optional[float]
+    nEQR: Optional[float]
```
haakondr, line 599: are all these values optional in all cases? Or is it like, if EQR is set, nEQR should also be set?

src/odm2_postgres_api/routes/begroing_routes.py
```diff
@@ -231,43 +232,49 @@ async def post_indices(
...
-            unitsid=dimensionless_unit.unitsid,
-            processinglevelid=processing_level.processinglevelid,
-            valuecount=1,
-            statuscv="Complete",
-            sampledmediumcv="Organism",
-            dataqualitycodes=[],
-        )
+    for prefix in new_indices.indices:
+        new_values: BasicIndexingUnit = new_indices.indices[prefix]
+        for suffix in new_values:
```
haakondr, line 238: This can be rewritten to a oneliner:

```
for prefix, new_values in new_indices.indices.items():
  ...
```

src/odm2_postgres_api/routes/begroing_routes.py
```diff
@@ -231,43 +232,49 @@ async def post_indices(
...
+            value = new_values.get(suffix)
+            if value:
```
haakondr, line 240: this if-check is truthy, so I think if value is 0 then the if check is false. Perhaps not what we want?

src/odm2_postgres_api/schemas/schemas.py
```diff
@@ -3,6 +3,7 @@
 from typing import Optional, List, Tuple, Dict, Union
 
 import shapely.wkt
+from mypy.suggestions import TypedDict
```
haakondr, line 6: unused import. I think if we'd use TypedDict, we'd use the following import:

```
from typing import TypedDict
```

haakondr (approved):
Looks good I think. I think the if check for truthy values needs to be considered. Could be ok, but not sure

src/odm2_postgres_api/schemas/schemas.py
roryscot: Optional in all cases, as far as I know.

tests/test_begroing.py
```diff
@@ -41,12 +41,18 @@ async def test_post_new_indices(db_conn):
     created_sampling_feature = await post_sampling_features(SamplingFeaturesCreate(**sf), db_conn)
     assert created_sampling_feature.samplingfeatureid > 0
 
-    index_data = {
-        "project_ids": [project_response.directiveid],
-        "station_uuid": created_sampling_feature.samplingfeatureuuid,
-        "date": "2020-09-01T00:00:00.000Z",
-        "indices": [{"indexType": "PIT EQR", "indexValue": "11"}],
-    }
+    index_data = (
```
RmStorm, line 44: The old `index_data` was a dict while this a tuple. This breaks the next line! This should remain a dict.

src/odm2_postgres_api/routes/begroing_routes.py
RmStorm: Yeah, if these submodels are `Dict` you can do it like that:
```python
    for prefix, indexing_unit in beg_thing.indices.items():
        for suffix, value in indexing_unit.items():
            print(prefix, suffix, value)
```
If the submodels are also converted to pydantic `BaseModel` (as they should) you can drop the `.items()` call.
```python
    for prefix, indexing_unit in beg_thing.indices:
        for suffix, value in indexing_unit:
            print(prefix, suffix, value)
```

src/odm2_postgres_api/routes/begroing_routes.py
RmStorm: I agree, it should explicitly check `value is None`.

src/odm2_postgres_api/routes/begroing_routes.py
```diff
@@ -208,10 +209,10 @@ async def post_indices(
         isactionlead=True,
         methodcode="begroing_6",  # code begroing_6 = Begroing Index Calculation
         actiontypecv="Derivation",
-        begindatetime=new_index.date,
+        begindatetime=new_indices.date,
         begindatetimeutcoffset=0,
         equipmentids=[],
-        directiveids=new_index.project_ids,
+        directiveids=new_indices.project_ids,
```
RmStorm, line 215: Pretty much this whole endpoint starting from the `post_actions()` call must be wrapped in a transaction like so:
```python
async with connection.transaction():
```

RmStorm (changes requested):
Hey Rory, sorry for the delayed review. There are three problems with the code as is. I think the schemas should really derive from pydantic `BaseModel` instead of from `Dict` since this is what fastApi expects afaik. Also, the test is broken since a variable was changed from dict to tuple. The test should be fixed. Lastly, the post index endpoint must be wrapped in a transaction!

src/odm2_postgres_api/schemas/schemas.py
```diff
@@ -593,16 +594,22 @@ class BegroingResult(BegroingResultCreate):
     personid: int
 
 
-class IndicesInfo(BaseModel):
-    indexType: str
-    indexValue: float
+class BasicIndexingUnit(Dict):
```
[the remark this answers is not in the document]
roryscot: Ya, I was getting weird linting errors on the routes

tests/test_begroing.py
roryscot: This was something that Black changed on save

src/odm2_postgres_api/schemas/schemas.py
roryscot: src/odm2_postgres_api/routes/begroing_routes.py:237: error: Value of type "IndexingInfo" is not indexable
src/odm2_postgres_api/routes/begroing_routes.py:239: error: "BasicIndexingUnit" has no attribute "get"
src/odm2_postgres_api/routes/begroing_routes.py:241: error: No overload variant of "__add__" of "tuple" matches argument type "str"
src/odm2_postgres_api/routes/begroing_routes.py:241: note: Possible overload variants:
src/odm2_postgres_api/routes/begroing_routes.py:241: note:     def __add__(self, Tuple[Any, ...]) -> Tuple[Any, ...]

PR #102, opened 2021-01-11
Mass Spec: remove site creation and make date and site optional for sample

jemmima:
This change is to enable mass spec ui to register samples (LabWare codes) in odm2 without needing site or date metadata.
 2 changes:
- remove the ability for mass-spec to create sites
- make sample creation possible with only labware code, no site and sample. This is necessary for external customers for whom the sample collection metadata are secret.

Goes with:

commit https://github.com/NIVANorge/nivacloud-mass-spectrometry/commit/80ff37d754b8da4b6ba268b8408bc531cec2047b
and
pr https://github.com/NIVANorge/mass-spec-ui/pull/25


src/odm2_postgres_api/queries/mass_spec_select_queries.py
```diff
@@ -222,50 +222,47 @@ async def register_sample(conn: asyncpg.connection, data: schemas.MsCreateSample
     async with conn.transaction():
         samplingfeatureid = await get_samplingfeatureid_from_samplingfeaturecode(conn, data.samplingfeaturecode)
         if samplingfeatureid is None:
+            if data.parent_samplingfeatureid is None:
+                relatedsamplingfeatures = []
+            else:
+                relatedsamplingfeatures = [(data.parent_samplingfeatureid, "Was collected at")]
+
```
jemmima, line 229: This handles the case where a sample is being created without a parent site

src/odm2_postgres_api/queries/mass_spec_select_queries.py
```diff
@@ -222,50 +222,47 @@ async def register_sample(conn: asyncpg.connection, data: schemas.MsCreateSample
...
+                ms_sample_data = schemas.ActionsCreate(
+                    actiondescription="Collected water sample",
+                    actiontypecv="Specimen collection",
+                    methodcode="mass_spec:collect_sample",
+                    begindatetime=data.collection_time,
+                    isactionlead=True,
+                    sampling_features=[sampling_feature],
+                    affiliationid=1,
+                    begindatetimeutcoffset=0,
+                )
```
jemmima, line 258: This allows an action to be created for registering a water sample rather than collecting it if no date and parent site are provided. Feedback on how to better work this into the data model would be great!

src/odm2_postgres_api/queries/mass_spec_select_queries.py
```diff
@@ -222,50 +222,47 @@ async def register_sample(conn: asyncpg.connection, data: schemas.MsCreateSample
...
-                samplingfeaturetypecv="Site",
-                samplingfeaturedescription=data.description,
-                samplingfeaturename=data.samplingfeaturename,
-            )
-
-            completed_sample = await core_queries.create_sampling_feature(conn, sampling_feature)
-            samplingfeatureid = completed_sample.samplingfeatureid
-    return samplingfeatureid
-
-
```
jemmima, line 268: Removing site registration from mass-spec queries - this will no longer be something mass-spec can do.

src/odm2_postgres_api/routes/mass_spectrometry_routes.py
```diff
@@ -81,11 +80,6 @@ async def post_register_sample(data: schemas.MsCreateSample, connection=Depends(
     return await register_sample(connection, data)
 
 
-@router.post("/register_site")
-async def post_register_site(data: schemas.MsCreateSite, connection=Depends(api_pool_manager.get_conn)) -> int:
-    return await register_site(connection, data)
-
-
```
jemmima, line 88: removing mass-spec endpoint for registering sites

src/odm2_postgres_api/schemas/schemas.py
```diff
@@ -654,8 +654,8 @@ class MsReplicas(BaseModel):
 
 class MsCreateSample(BaseModel):
     samplingfeaturecode: str
-    parent_samplingfeatureid: int
-    collection_time: dt.datetime
+    parent_samplingfeatureid: Optional[int] = None
+    collection_time: Optional[dt.datetime] = None
```
jemmima, line 659: Making parent site and collection time optional when creating samples. This is important as sometimes samples will come from third parties and the site and collection time will be considered confidential/secret.

ZofiaRudjord (approved):
LGTM!

src/odm2_postgres_api/queries/mass_spec_select_queries.py
RmStorm, line 229: Now that I look at it again I think I prefer:
```python
relatedsamplingfeatures = []
if data.parent_samplingfeatureid is not None:
    relatedsamplingfeatures.append((data.parent_samplingfeatureid, "Was collected at"))
```
One line less ^^ but funky double parenthesis. Anyway, completely up to you.

src/odm2_postgres_api/queries/mass_spec_select_queries.py
RmStorm: I think this is clearer if you kinda 'split out' the only two fields that actually change instead of repeating the other 6 fields as well. key value unpacking offers a pretty compact way to do that:
```python
    if data.collection_time is None:
        variable_fields = {'actiondescription': "Registered water sample",
                           'begindatetime': datetime.now()}
    else:
        variable_fields = {'actiondescription': "Collected water sample",
                           'begindatetime': data.collection_time}
        **variable_fields,
       methodcode="mass_spec:collect_sample",
       isactionlead=True,
       sampling_features=[],
       affiliationid=1,
       begindatetimeutcoffset=0,
    )
```

RmStorm (approved):
Looks good to me! I don't quite know why site registration is no longer needed but I assume you and @ZofiaRudjord  discussed this. I also came up with some code snippets that are a little more concise. You can choose if you want to use them or not. It does not make much difference so you can ignore them freely :wink:

src/odm2_postgres_api/queries/mass_spec_select_queries.py
```diff
@@ -222,50 +222,47 @@ async def register_sample(conn: asyncpg.connection, data: schemas.MsCreateSample
...
-                sampling_features=[sampling_feature],
-                affiliationid=1,
-                begindatetimeutcoffset=0,
-            )
+            if data.collection_time is None:
+                ms_sample_data = schemas.ActionsCreate(
+                    actiondescription="Registered water sample",
+                    actiontypecv="Specimen collection",
+                    methodcode="mass_spec:collect_sample",
+                    begindatetime=datetime.now(),
```
haakondr, line 242: @RmStorm should we specify utc timezone when declaring timestamps?

src/odm2_postgres_api/queries/mass_spec_select_queries.py
haakondr, line 258: I am thinking that the other "mode" with registering sample including site and collection time will not be used, so we could consider to remove this functionality entirely. 

If we setup a sync labware/nivabasen -> cloud the samples will be automatically created. Maybe registered using this endpoint, but the workflow will surely be different so I think it would be best to just change this endpoint to only accept labware code.

Please correct me if I make wrong assumptions here :)

src/odm2_postgres_api/queries/mass_spec_select_queries.py
haakondr: @RmStorm @jemmima  is this not currently in use by begroing app?

haakondr: never mind: I see that begroing is using POST /sampling_features

src/odm2_postgres_api/schemas/schemas.py
haakondr: I think instead of making this optional, we can have two different types and two different endpoints, differentiating them by url/path.

Also see my comment above about whether we should remove the functionality for storing samples with date/site. I think this will potentially end up as dead code, only adding to the complexity of the API

haakondr (approved):

src/odm2_postgres_api/queries/mass_spec_select_queries.py
ZofiaRudjord: I think the way of thinking here was that if a sample does not have a labware code it can be registered here and additional info regarding collection can be registered. 

We plan to discuss metadata registration with Malcolm and Pepo during app presentation meeting on Thursday.

jemmima: Nice! Thanks!

src/odm2_postgres_api/queries/mass_spec_select_queries.py
jemmima: I would argue that no other apps should be using the mass_spec_select_queries. We should have a place for generic/shared queries but mass spec needs to be able to maintain it's own queries without worrying other apps will be impacted.

src/odm2_postgres_api/schemas/schemas.py
jemmima: Re: the two types and two endpoints - should I make them in shemas.py? Or have the optional one somewhere mass-spec specific? 

Sorry for the delay I've been off sick! I've only removed the mass spec specific logic, not that used by other apps. I understood that we don't want this functionality in mass spec, at least not for a while. We will have it in the history so can add it back if that changes later. What do you think?

src/odm2_postgres_api/queries/mass_spec_select_queries.py
jemmima: Thanks for the tip! :)

src/odm2_postgres_api/queries/mass_spec_select_queries.py
jemmima: @haakondr are we not only removing it from mass_spec? Are there not likely to be other use cases with other apps that would want associated site and collection date? Mass spec must have wanted it for it to be there in the beginning, but as Zofia says, we can ask them next week.

src/odm2_postgres_api/queries/mass_spec_select_queries.py
RmStorm: yeah that's probably better! Good catch.

jemmima: I had a quick google - this should be datetime.utcnow() correct?

src/odm2_postgres_api/queries/mass_spec_select_queries.py
jemmima: I'm going to merge this as it is and change it if required after we meet with Malcolm and Pepo next week.

PR #103, opened 2021-01-18
Feature/#10 filter samples

jemmima:
Add an annotation for mass spec samples to annotations table and create samplingfeatureannotations for each new sample registered by mass-spec-api. To follow I will modify the api query from mass-spec-ui to fetch only the mass spec samples. This pr is independent so can be merged before that is done.


src/odm2_postgres_api/queries/mass_spec_select_queries.py
```diff
@@ -243,6 +243,13 @@ async def register_sample(conn: asyncpg.connection, data: schemas.MsCreateSample
 
             completed_sample = await core_queries.do_action(conn, ms_sample_data)
             samplingfeatureid = completed_sample.sampling_features[0].samplingfeatureid
+            sampling_feature_annotation_create = schemas.SamplingFeatureAnnotationCreate(
+                samplingfeatureid=samplingfeatureid,
+                annotationid=15,
+                annotationtypecv="Specimen annotation",
+                annotationtext="Mass spectrometry sample",
+            )
+            await core_queries.create_sampling_feature_annotation(conn, sampling_feature_annotation_create)
```
jemmima, line 252: This is the main part I am unsure of - I tested it and it works, but I'm not sure if this is good practice or the right place for it. Should it be in register_sample or somewhere else? 🤔

ZofiaRudjord: You already created annotation in mass_spec_metadata and now you want to link it to the samplingfeature you create. I think you can re-use the annotation field in SamplingFeaturesCreate. You can get the annotation id via fetch_row.

ZofiaRudjord (changes requested):

src/odm2_postgres_api/queries/storage.py
```diff
@@ -140,3 +142,10 @@ async def save_methods(connection, method: MethodsCreate) -> Methods:
     if existing:
         return existing
     return await core_queries.insert_method(connection, method)
+
+
+async def save_annotations(connection, annotation: AnnotationsCreate) -> Annotations:
+    existing = await find_row(connection, "annotations", "annotationtypecv", annotation.annotationtypecv, Annotations)
```
RmStorm, line 148: I don't think this should try to find a row based on `annotationtypecv` it should look at the `annotationtext` instead. That is still not 100% solid since annotationtext doesn't have a unique constraint. It should however be solid enough to find the annotation with text: "Mass spectrometry sample".

src/odm2_postgres_api/queries/mass_spec_select_queries.py
RmStorm: yeah I agree, the `annotationid` can be added directly to the `SamplingFeaturesCreate` on line 226 like so:
```python
            mass_spec_annotation = await find_row(conn, "annotations", "annotationtext",
                                                  "Mass spectrometry sample", schemas.Annotations)
            sampling_feature = schemas.SamplingFeaturesCreate(
                samplingfeatureuuid=uuid.uuid4(),
                samplingfeaturecode=data.samplingfeaturecode,
                samplingfeaturetypecv="Specimen",
                relatedsamplingfeatures=[(data.parent_samplingfeatureid, "Was collected at")],
                annotations=[mass_spec_annotation.annotationid]
            )
```

RmStorm: Fetching the `annotationid` like that instead of hardcoding `15` is more solid I think!

RmStorm (changes requested):
I agree with @ZofiaRudjord that the annotationid should be fetched instead of hardcoded. It can also be added directly to the samplingfeaturecreate query instead of being it's own query.

Also, in the save_annotations function it's better to use find_row on the annotationtext instead of the annotationtypecv.

src/odm2_postgres_api/queries/mass_spec_select_queries.py
jemmima: 🤦‍♀️  of course!! How's this @ZofiaRudjord ?

```
async def register_sample(conn: asyncpg.connection, data: schemas.MsCreateSample) -> int:
    """
    This function registers the main sample collected in samplingfeature table.
    """
        samplingfeatureid = await get_samplingfeatureid_from_samplingfeaturecode(conn, data.samplingfeaturecode)
        if samplingfeatureid is None:

            annotationid = await conn.fetch(
                "select annotationid from annotations a "
                "where annotationtext = 'Mass spectrometry sample'"
            )

            sampling_feature = schemas.SamplingFeaturesCreate(
                samplingfeatureuuid=uuid.uuid4(),
                samplingfeaturecode=data.samplingfeaturecode,
                samplingfeaturetypecv="Specimen",
                annotations=[annotationid],
            )

                actiondescription="Registered water sample",
                begindatetime=datetime.utcnow(),
                methodcode="mass_spec:collect_sample",
                isactionlead=True,
                sampling_features=[sampling_feature],
                affiliationid=1,
                begindatetimeutcoffset=0,
            )

            completed_sample = await core_queries.do_action(conn, ms_sample_data)
            samplingfeatureid = completed_sample.sampling_features[0].samplingfeatureid

    return samplingfeatureid
```

ZofiaRudjord: Doing it this way will also create the bridge between annotation_id created during metadata startup and 
sampling_feature_id

jemmima: Ah you beat me to it! I just pushed this but my find_row is different 🤔. Should I change to what you have here @RmStorm ?

src/odm2_postgres_api/queries/storage.py
jemmima: Yes I totally agree!!

jemmima: Changed in ef2f5ff. Is this ok now?

jemmima: Changed in ef2f5ff, is this ok now?

src/odm2_postgres_api/queries/mass_spec_select_queries.py
jemmima: > Doing it this way will also create the bridge between annotation_id created during metadata startup and
> sampling_feature_id

The way I had it did create the bridge because create_sampling_feature_annotation does this and checks if the annotation exists already, but it was hardcoding the annotation_id, which I suppose could be different for every metadata startup? Either way I totally agree that the find_row is better for several reasons!

src/odm2_postgres_api/queries/storage.py
ZofiaRudjord: Looks good!

src/odm2_postgres_api/queries/mass_spec_select_queries.py
```diff
@@ -217,17 +218,19 @@ async def register_replicas(conn: asyncpg.connection, data: schemas.MsCreateRepl
...
 async def register_sample(conn: asyncpg.connection, data: schemas.MsCreateSample) -> int:
     """
-    This function registers the main sample collected at a given Site in samplingfeature table.
+    This function registers the main sample collected in samplingfeature table.
     """
     async with conn.transaction():
         samplingfeatureid = await get_samplingfeatureid_from_samplingfeaturecode(conn, data.samplingfeaturecode)
         if samplingfeatureid is None:
 
+            annotationid = await find_row(conn, "annotations", "annotationtext", "Mass spectrometry sample", schemas.Annotations)
```
RmStorm, line 227: I think find_row returns the entire annotation, not just the annotationid..

jemmima: hmm 🤔  that makes sense but why does this work? 😅

jemmima: Changed in 73816a5!

RmStorm (approved):
LGTM! Maybe there should be a note on the `annotation_text` `find_row` bit that there is a potential bug hiding there due to lack of a unique constraint.

Should we add a joint unique constraint on the `annotationtype` and the `annotationtext`? These are the only two columns that are obligatory in the annotations table. A joint unique constraint would allow us to query for that without ever hitting a duplicate and I think it might make sense anyways :thinking: @ZofiaRudjord @jemmima what do you think?

src/odm2_postgres_api/queries/mass_spec_select_queries.py
RmStorm: I think it worked because the Annotation type is derived from the AnnotationCreate type which was also valid instead of an id for identifying the annotation...

jemmima:
> LGTM! Maybe there should be a note on the `annotation_text` `find_row` bit that there is a potential bug hiding there due to lack of a unique constraint.
> 
> Should we add a joint unique constraint on the `annotationtype` and the `annotationtext`? These are the only two columns that are obligatory in the annotations table. A joint unique constraint would allow us to query for that without ever hitting a duplicate and I think it might make sense anyways 🤔 @ZofiaRudjord @jemmima what do you think?

We would have to change how the result annotations are working now because they are not unique:

jemmima:
I have added annotations for replica registration as well following a conversation on slack with @ZofiaRudjord and @RmStorm be82c37

haakondr (approved):

src/odm2_postgres_api/queries/mass_spec_select_queries.py
```diff
@@ -127,20 +128,22 @@ async def register_replicas(conn: asyncpg.connection, data: schemas.MsCreateRepl
     """
     async with conn.transaction():
         action_time = datetime.now()
+        parameters_annotation_id = schemas.AnnotationsCreate(
```
RmStorm, line 131: Can this variable be renamed to `parameters_annotation` since it's not and id but the entire annotation?

RmStorm (approved):
Looks good to me! I have made a final comment about changing a variable name :grin: It's optional so feel free to ignore it!

src/odm2_postgres_api/queries/mass_spec_select_queries.py
jemmima: Yes! I didn't actually change this part and I didn't notice it was incorrect.

src/odm2_postgres_api/queries/mass_spec_select_queries.py
```diff
@@ -127,20 +128,22 @@ async def register_replicas(conn: asyncpg.connection, data: schemas.MsCreateRepl
...
-            annotations=[
-                schemas.AnnotationsCreate(
-                    annotationjson=data.samplingfeatureannotationjson,
-                    annotationdatetime=action_time,
-                    annotationtypecv="Specimen annotation",
-                    annotationtext="Processing parameters",
-                    annotationutcoffset=0,
-                )
-            ],
+            annotations=[parameters_annotation.annotationid, ms_annotation.annotationid],
```
RmStorm, line 146: This list `annotations` accepts both integers or `AnnotationsCreate` objects. So it should be:
```python
annotations=[parameters_annotation, ms_annotation.annotationid]
```
and that will fix the failing test ^^

ZofiaRudjord: we are on it!

src/odm2_postgres_api/queries/mass_spec_select_queries.py
jemmima: It wasn't incorrect! I was confused. Fixed now in

src/odm2_postgres_api/queries/mass_spec_select_queries.py
jemmima: Thanks! Fixed in 30d3361

ZofiaRudjord (approved):
LGTM

PR #107, opened 2021-02-16
Added project requirement to MsCreateSample, updated register_sample …

ZofiaRudjord:
As discussed with @jemmima we need a concept of a project in mass spec processing in order to have a higher abstraction to group samples into "signal" and "blanks". 

I Added project requirement to MsCreateSample , updated register_sample function to link to the project, added mass_spec:test project to mass spec metadata. 

This change is not backward compatible, as directivedescription is now required when registering sample.
This will break both mass-spec-ui as well as ms_init ( and the integration test).


src/odm2_postgres_api/queries/storage.py
```diff
@@ -149,3 +151,12 @@ async def save_annotations(connection, annotation: AnnotationsCreate) -> Annotat
     if existing:
         return existing
     return await core_queries.create_or_parse_annotations(connection, [annotation])
+
+
+async def save_directives(connection, directive: DirectivesCreate) -> Directive:
+    existing = await find_row(
+        connection, "directives", "directivedescription", directive.directivedescription, Directive
+    )
```
ZofiaRudjord, line 159: directivedescription is not a unique constraint but it should be I think. What is the best practise to impose such constraint in the API @haakondr

haakondr (approved):
looks ok, but I am not sold on using a description field as a grouping mechanism. Should we consider to use annotations ? Or link them by method?

src/odm2_postgres_api/metadata_init/data/mass_spec/mass_spec_metadata.py
```diff
@@ -21,6 +22,16 @@ def mass_spec_annotations() -> List[AnnotationsCreate]:
...
 
 
+def mass_spec_directives() -> List[DirectivesCreate]:
+    mass_spec_directives_list = [
+        {
+            "directivetypecv": "Project",
+            "directivedescription": "mass_spec:test",
+        }
+    ]
+    return [DirectivesCreate(**d) for d in mass_spec_directives_list]
```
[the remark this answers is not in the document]
ZofiaRudjord: yes

src/odm2_postgres_api/queries/storage.py
haakondr: ALTER TABLE ODM2.directives ADD UNIQUE (directivedescription);

haakondr:
may consider to introduce directives.name column.

ZofiaRudjord:
I added the directivename column to store the project name. This column has the uniqueness constraint.
Grouping into blank/subject will be done via annotations.

src/odm2_postgres_api/schemas/schemas.py
```diff
@@ -652,8 +654,18 @@ class MsReplicas(BaseModel):
     results: Results
 
 
+class MsSampleTypeEnum(str, Enum):
+    subject = "subject"
+    blank = "blank"
```
jemmima, line 659: Are the blanks called blank replicas or just blanks? If they're only referred to as blanks can we call the subject ones `replicas` instead? Then it would be replicas and blanks. If the blanks are blank replicas then the others should be called subject replicas.

ZofiaRudjord: Technically sample is "blank", but replicas of blank samples are for short often referred to as blanks. I think it is useful to keep the distinction we have that replicas are binary files which come from mass spec, as oppose to samples which are water samples.

jemmima (approved):
LGTM!

PR #108, opened 2021-02-16
odm2 to master

ZofiaRudjord:
Looks like we have not merged to master for a long time, any reasons for that @haakondr @jemmima  ?


k8s/README.md
```diff
@@ -35,3 +35,13 @@ make 4 secrets, 2 for nivates and 2 for nivaprod. Apply them using:
...
+
+## Aquamonitor API user
+
+this can be the same for nivatest/nivaprod:
+
+```
+echo -n "longsecretpassword" > password
+kubectl create secret generic  aquamonitor-api-user --from-file=password --from-literal='username=nivacloud'
+rm password
+```
```
haakondr, line 47: note to self: this should really only be setup in production. Not in use yet but will be once begroing is up and running

k8s/deployments/deployment-odm2.tmpl.yaml
```diff
@@ -48,7 +48,7 @@ spec:
       containers:
       - name: timescale-odm2
         # Remember to change the image version in both nivacloud/docker-compose.yml and the odm2/docker-compose.yml
-        image: timescale/timescaledb-postgis:1.4.2-pg11
+        image: timescale/timescaledb-postgis:1.7.4-pg12
```
haakondr, line 51: @ZofiaRudjord here we are upgrading postgres version, that will probably require some manual follow up when we merge to master. I don't recall all the details, but I think kubernetes statefulsets don't get replaced by just changing the image. Need to explicitly delete the pod for it to take effect.

k8s/deployments/deployment-odm2.tmpl.yaml
```diff
@@ -75,7 +75,7 @@ spec:
               name: odm2-postgres-password
               key: password
         - name: PGDATA
-          value: /var/lib/postgresql/data/pg11
+          value: /var/lib/postgresql/data/pg12
```
haakondr, line 78: not so sure about this change. I think I did this, but here we are changing the location of our data. For odm2 prod, do we have any data worth keeping or can we go for a full wipe?

src/odm2_postgres_api/aquamonitor/aquamonitor_api_types.py
```diff
@@ -0,0 +1,121 @@
...
+    IdentifiedBy: Optional[str] = None
+    Remark: Optional[str] = None
+
+
+class BegroingObservationCargo(BegroingObservationCargoCreate):
+    Id: int
+
+
+class AquamonitorAPIError(HTTPException):
+    pass
```
haakondr, line 121: plan to separate this out as a python package wrapping aquamonitor API in the future

src/odm2_postgres_api/db_init/ODM2_for_PostgreSQL.sql
```diff
@@ -2319,4 +2320,49 @@ create index TrackResultValues_resultid_time_idx
...
+
+ALTER TABLE ODM2.personexternalidentifiers
+    ADD UNIQUE (personexternalidentifier, externalidentifiersystemid);
+
+ALTER TABLE ODM2.affiliations ADD UNIQUE (primaryemail);
+
+ALTER TABLE ODM2.externalidentifiersystems ADD UNIQUE (externalidentifiersystemname);
+ALTER TABLE ODM2.units ADD UNIQUE (unitstypecv, unitsabbreviation);
+
+ALTER TABLE ODM2.samplingfeatures ADD UNIQUE (samplingfeatureuuid);
```
haakondr, line 2335: some of these new unique constraints may fail when adding to a new environment (production), need to look out for that

src/odm2_postgres_api/metadata_init/data/general/people.py
```diff
@@ -0,0 +1,14 @@
+from odm2_postgres_api.schemas.schemas import PeopleAffiliationCreate
+
+
+def unknown_person(org_id: int) -> PeopleAffiliationCreate:
```
haakondr, line 4: Not sure if this is used anymore, cannot see any references to it. I think the intention here is to use it in cases where we don't have sufficient data about who did something. Instead of setting the column to nullable, we use this.

That being said, it is a bit hacky and should definitely have a valid email at least.

src/odm2_postgres_api/queries/core_queries.py
```diff
@@ -1,67 +1,213 @@
...
+
+async def find_row(conn: asyncpg.connection, table: str, id_column: str, identifier, model, raise_if_none=False):
+    row = await conn.fetchrow(f"SELECT * FROM {table} WHERE {id_column}=$1", identifier)
+    if row is None and raise_if_none:
+        raise HTTPException(status_code=422, detail=f"Item in table {table} with '{id_column}={identifier}' not found")
+    return model(**row) if row else None
+
+
+# TODO: we may want to extend this further by including organization ++
+async def find_person_by_external_id(conn: asyncpg.connection, external_system, ext_id: str) -> PersonExtended:
```
haakondr, line 59: I started adding all users from on-premise active directory into ODM2 in order to have a link between 3-letter usernames and full user objects.

I think it works, but need to think a bit if we want to have this information in ODM2. GDPR may play a role

src/odm2_postgres_api/queries/core_queries.py
```diff
@@ -79,122 +225,243 @@ async def create_sampling_feature(conn: asyncpg.connection, sampling_feature: sc
...
-                                  "(SELECT featureactionid FROM results WHERE resultid = $1)", track_result.resultid)
+        row = await conn.fetchrow(
+            "SELECT samplingfeatureid FROM featureactions WHERE featureactionid = "
+            "(SELECT featureactionid FROM results WHERE resultid = $1)",
+            track_result.resultid,
+        )
 
         if row["samplingfeatureid"] != track_result.samplingfeatureid:
-            raise ValueError('THIS IS ALLL WROOONGNGNGNGNGNNG')
+            raise ValueError("THIS IS ALLL WROOONGNGNGNGNGNNG")
```
haakondr, line 372: hm

src/odm2_postgres_api/queries/mass_spec_select_queries.py
```diff
@@ -0,0 +1,325 @@
...
+            annotations=[parameters_annotation, ms_annotation.annotationid],
+        )
+
+        fractionate_sample_data = schemas.ActionsCreate(
+            actiondescription="Fractionate water sample",
+            actiontypecv="Specimen fractionation",
+            methodcode="mass_spec:fractionate_sample",
+            begindatetime=action_time,
+            sampling_features=[sampling_feature],
+            affiliationid=1,
```
haakondr, line 156: @ZofiaRudjord @jemmima affiliationid should not be hardcoded. Should be based on the logged in user. 

Have a look at https://github.com/NIVANorge/odm2-postgres-api/blob/dev/src/odm2_postgres_api/routes/begroing_routes.py#L56-L58

In short, niva-port takes the user token and adds its content as a base64 encoded http header to all port apps on internal http requests. This can be read by the odm2 API and used to get the affiliation id:

```
@router.post("/begroing_result", response_model=schemas.BegroingResult)
async def post_begroing_result(
    begroing_result: schemas.BegroingResultCreate,
    connection=Depends(api_pool_manager.get_conn),
    niva_user: str = Header(None),
):
    user = await create_or_get_user(connection, niva_user)
    ....
    
     data_action = schemas.ActionsCreate(
                affiliationid=user.affiliationid,
                ....
                )
```

src/odm2_postgres_api/queries/mass_spec_select_queries.py
```diff
@@ -0,0 +1,325 @@
...
+            valuecount=0,
+            resulttypecv="Measurement",
+            sampledmediumcv="Liquid aqueous",
+            annotations=[resultannotation],
+            resultdatetime=action_time,
+            validdatetime=action_time,
+            resultdatetimeutcoffset=0,
+            validdatetimeutcoffset=0,
+            actionid=completed_ran_mass_spec.actionid,
+            resultuuid=uuid.uuid4(),
```
haakondr, line 211: I guess this is a matter of preference, but I think it could be a good practice to generate uuids on the clientside.

Depends on the use case, but imagine if we build a user interface with a list of items. We may add a new item to the list, then fill in some values. When we are done, we save it by doing a POST request to the API.

React and other frameworks typically require you to assign a key to the item so that the framework can keep track of changes to the list: https://reactjs.org/docs/lists-and-keys.html#keys

If we generate the uuid clientside, we can set this as the key in the UI and update it based on the response (if there are any fields set serverside). This way, the UI framework can preserve order and know which item to update based on the response.

All this may not be relevant for this change, but thought I'd mention it

src/odm2_postgres_api/routes/begroing_routes.py
```diff
@@ -0,0 +1,282 @@
...
+    "HBI2": 14,
+    "PIT EQR": 15,
+    "AIP EQR": 16,
+    "HBI EQR": 17,
+    "HBI2 EQR": 18,
+    "PIT nEQR": 19,
+    "AIP nEQR": 20,
+    "HBI nEQR": 21,
+    "HBI2 nEQR": 22,
+}
```
haakondr, line 47: hmm, this may not work. hardcoding to sequential id's is not a good idea. Will need to test this before we get production data of begroing in odm2

src/odm2_postgres_api/routes/begroing_routes.py
```diff
@@ -0,0 +1,282 @@
...
+                        timeaggregationinterval=0,
+                        timeaggregationintervalunitsid=seconds_unit.unitsid,
+                        datavalue=value,
+                        valuedatetime=data.date,
+                        valuedatetimeutcoffset=0,
+                    )
+
+                    await post_measurement_results(data_measurement_result, connection)
+
+    return schemas.BegroingIndices(personid=user.personid, **data.dict())
```
haakondr, line 282: the begroing functionality will need thorough testing before actual use, so I don't think there is a big risk of having it in production

src/odm2_postgres_api/routes/fish_rfid/fish_rfid_repository.py
```diff
@@ -0,0 +1,91 @@
...
+                fish_sampling_feature=fish_sf.samplingfeatureuuid,
+                station_sampling_feature=station_sf.samplingfeatureuuid,
+            )
+        )
+
+    return FishObservationResponse(
+        observations=stored_observations,
+        fish_sampling_features=fish_sampling_features,
+        station_sampling_features=station_sampling_features,
+    )
```
haakondr, line 91: same goes for fish_rfid stuff. James is using this to some degree now, but will need some more testing

src/odm2_postgres_api/utils/google_cloud_utils.py
```diff
@@ -74,7 +75,7 @@ def put_csv_to_bucket(csv_data):
         for row in csv_data:
             writer.writerow(row)
 
-        content = csv_file.getvalue().encode('utf-8')
+        content = csv_file.getvalue().encode("utf-8")
 
     with BytesIO(initial_bytes=content) as csv_file:
         upload_blob_wrapper(bucket_name, file_name, csv_file)
```
haakondr, line 81: I think all this can be deleted, was a temp solution

haakondr (approved):
this should have been done a long time a ago. 

A good mix of things that are not yet in use, some that are needed (mass spec?). Added some comments.

Will need to pay close attention to database upgrade, also need to figure out if any of the data in nivaprod are worth keeping. The postgres deployment changes data directory so that basically means that it sets up a new database. This is probably not needed

PR #110, opened 2021-03-04
Added mass_spec filter 

ZofiaRudjord:
Added mass_spec filter methods to metadata init, extended filtering in find_result_annotationlink to check for replicas and blanks if filter required. Modified schemas accordingly. These changes are necessary to solve #21 in nivacloud-mass-spectrometry


src/odm2_postgres_api/queries/mass_spec_select_queries.py
```diff
@@ -41,6 +44,13 @@ async def find_result_annotationlink(
                 # methodcode defaults to None because of the Pydantic model
                 if methodcode and parameters["Methods"][method_code_type] != methodcode:
                     valid = False
+            if data.annotationcode == "fd_filtered":
+                if filter != parameters["filter"]:
+                    valid = False
+                if (filter == "replicas" or filter == "replicas_and_blanks") and replicas != parameters["replicas"]:
+                    valid = False
+                if (filter == "blanks" or filter == "replicas_and_blanks") and blanks != parameters["blanks"]:
```
haakondr, line 52: could use the MsFilterEnum to refer to "replicas_and_banks" like this:

```
MsFilterEnum.replica_and_blank
```

would yield the same result, but you gain IDE support when refactoring, viewing usage of enum, etc

haakondr (approved):

jemmima (approved):
LGTM!

PR #119, opened 2021-05-10
Metadata and routes for handling data of one time series (OxygenSat1_Avg) measured in Langtiern.

ZofiaRudjord:
The aim of this code is to create minimal set of metadata and routes to be able to post data to timeseriesresultvalue table
in odm2 by nivacloud-logger app: ingestor.

I added metadata for one timeseries: OxySat1_Avg measured in langtjern. For now this is just timeseries registartion. Additional metadata related to equipment deployed and linking to the column in ingested logger datafile will be added at a later stage. This PR is already quite extensive.

In addition I created 2 routes: one for getting resultid of time series of interest and the second to upsert timeseriesvalues.
I tested it on localhost:8701 and both routes function as expected, results are also inserted in my local database.


src/odm2_postgres_api/metadata_init/data/loggers/loggers_metadata.py
```diff
@@ -0,0 +1,63 @@
...
+
+from odm2_postgres_api.schemas.schemas import (
+    SamplingFeaturesCreate,
+    VariablesCreate,
+    EquipmentModelCreate,
+    LoggersTimeSeriesCreate,
+)
+
+
+def loggers_sampling_features() -> List[SamplingFeaturesCreate]:
```
haakondr, line 13: sampling features is a very abstract term. Could perhaps rename this to logger_stations or something? I assume that is what it is, in this context (?)

src/odm2_postgres_api/metadata_init/data/loggers/loggers_metadata.py
```diff
@@ -0,0 +1,63 @@
...
+            "isinstrument": True,
+        },
+    ]
+    return [EquipmentModelCreate(**em) for em in equipment_models]
+
+
+def loggers_timeseries_result() -> List[LoggersTimeSeriesCreate]:
+    timeseries = [
+        {
+            "samplingfeaturecode": "Langtjern_boye",
```
haakondr, line 56: This may be a good time to start thinking of a station code naming convention to be used in our new system. Not sure what would be the best format, but station codes and names are a human readable representation of a location.

One suggestion is that station codes and names should say something about _where_ they are and not _what_ they are used for. I.e. "langtjern outlet south" or something. I don't think we need to decide on this now, but I think the overall user experience would benefit from a good naming convention.

haakondr: PS: Not sure if "outlet" is really what I think it means in english. Was looking for the english word for "inn/ut-os", where water flows in/out of a lake

src/odm2_postgres_api/queries/core_queries.py
```diff
@@ -448,6 +475,14 @@ async def upsert_categorical_result(conn: asyncpg.connection, categorical_result
         )
 
 
+async def upsert_timeseries_resultvalues(
```
haakondr, line 478: I don't think the function make_sql_query creates an upsert statement (insert or update). It seems to only do insert. 

not sure if this is a problem, but may want to rename the function to `insert_timeseries_resultvalues` if that is the case

src/odm2_postgres_api/queries/loggers_queries.py
```diff
@@ -0,0 +1,98 @@
...
+        existing = await core_queries.find_timeseries(conn, data)
+        if existing:
+            return existing
+        action_time = datetime.now()
+        # This will create a new action of deploying equipment every time we deploy new sensor.
+        # Shall we have just one action for all deployments?
+        deploy_sensor = schemas.ActionsCreate(
+            actiontypecv="Instrument deployment",
+            methodcode="000",
+            affiliationid=1,
```
haakondr, line 28: maybe make some sort of affiliation that this was created by the metadata script? not sure how well this fits into odm2, but hardcoding affiliationid to 1 is a bit fragile. Not sure what the best solution would be here though, kinda makes sense to either set this field to nullable or create a dummy person/affiliation

src/odm2_postgres_api/queries/loggers_queries.py
```diff
@@ -0,0 +1,98 @@
...
+    It registers result and timeseries result of instrument deployment for a given variable,
+    and the corresponding action of deploying instrument.
+    """
+
+    async with conn.transaction():
+        # Check if exists to avoid creating duplicates
+        existing = await core_queries.find_timeseries(conn, data)
+        if existing:
+            return existing
+        action_time = datetime.now()
```
haakondr, line 22: should stick to the same timezone (utc) in all our date timestamps.

src/odm2_postgres_api/queries/loggers_queries.py
```diff
@@ -0,0 +1,98 @@
...
+        # This will create a new action of deploying equipment every time we deploy new sensor.
+        # Shall we have just one action for all deployments?
+        deploy_sensor = schemas.ActionsCreate(
+            actiontypecv="Instrument deployment",
+            methodcode="000",
+            affiliationid=1,
+            isactionlead=True,
+            begindatetime=datetime.utcnow(),
```
haakondr, line 30: could use action_time date created on line 22? will of course not make a difference, but nice to be sure that at least stuff like timezone is the same

src/odm2_postgres_api/routes/loggers_routes.py
```diff
@@ -0,0 +1,24 @@
...
+from odm2_postgres_api.queries.loggers_queries import upsert_timeseries
+
+router = APIRouter()
+
+
+@router.get("/get_timeseries_resultid/")
+async def get_timeseries_resultid(
+    data: schemas.LoggersTimeSeriesCreate,
+    connection=Depends(api_pool_manager.get_conn),
+) -> Optional[schemas.LoggersTimeSeriesId]:
```
haakondr, line 15: may want to throw an error if not found, Instead of making it optional? Kinda depends on the use case

src/odm2_postgres_api/routes/loggers_routes.py
```diff
@@ -0,0 +1,24 @@
...
+    return await find_timeseries(connection, data)
+
+
+@router.post("/time_series_result_values/")
+async def post_time_series_result_values(
+    data: List[schemas.LoggersTimeSeriesResultValuesCreate],
+    connection=Depends(api_pool_manager.get_conn),
+) -> List[schemas.LoggersTimeSeriesResultValues]:
+    return [await upsert_timeseries(connection, d) for d in data]
```
haakondr, line 24: here, a transaction is started in upsert_timeseries_resultvalues, but that is done per item in this list. I think we should wrap all writes in the same transaction. (could do it here)

src/odm2_postgres_api/metadata_init/populate_metadata.py
```diff
@@ -20,14 +22,19 @@
...
+)
 from odm2_postgres_api.metadata_init.data.general.controlled_vocabularies import (
     controlled_vocabularies,
 )
 from odm2_postgres_api.metadata_init.data.general.external_identifier_systems import (
     external_identifier_systems,
 )
 from odm2_postgres_api.metadata_init.data.general.methods import methods
-from odm2_postgres_api.metadata_init.data.general.organisations import niva_org
+from odm2_postgres_api.metadata_init.data.general.organisations import niva_org, campbell_org
```
knl88, line 37: campbell_org imported but never used, did you want to add it?

src/odm2_postgres_api/metadata_init/data/loggers/loggers_metadata.py
```diff
@@ -0,0 +1,63 @@
...
+from odm2_postgres_api.schemas.schemas import (
+    SamplingFeaturesCreate,
+    VariablesCreate,
+    EquipmentModelCreate,
+    LoggersTimeSeriesCreate,
+)
+
+
+def loggers_sampling_features() -> List[SamplingFeaturesCreate]:
+    # NOTE: how do we store coordinates of Site? sampling_feature_annotation?
```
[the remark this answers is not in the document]
knl88: I am still exploring odm2, but wouldn't https://odm2.github.io/ODM2/schemas/ODM2_Current/tables/ODM2SamplingFeatures_Sites.html make sence?

src/odm2_postgres_api/routes/loggers_routes.py
haakondr: also, depends on the usage of this endpoint, but I suspect that we will create duplicates if this is called twice? ref my comment about upsert not doing updates only inserts

src/odm2_postgres_api/metadata_init/data/loggers/loggers_metadata.py
ZofiaRudjord: Perhaps...I think it is good to stay consistent with naming and this is how this api is written. For me it is easier to read this code and get a quick understanding how different elements play together. How about you @knl88 ?

src/odm2_postgres_api/metadata_init/data/loggers/loggers_metadata.py
ZofiaRudjord: We discussed naming a bit with @jankarud . This is coming from data logger file header and also data logger filename.
Boye is exactly telling you what is it. It is neither outlet nor inlet which are separate data ( also taken in Langtjernet)

src/odm2_postgres_api/queries/core_queries.py
ZofiaRudjord: I think we want it to upsert I will investigate.

src/odm2_postgres_api/metadata_init/populate_metadata.py
ZofiaRudjord: I implemented instrument metadata and then I decided to remove it from this PR and forgot to delete campbell_org. I will delete it.

src/odm2_postgres_api/routes/loggers_routes.py
knl88: Guess this is a bit nitpicking, so ignore my comment freely;) I think the get_timeseries_resultid is a bit long, with the method being GET, you don't need the get in the route. Also from a user perspective I do like using the plural form where possible and return a list that is empty for no match. I also take the opportinity to complain about snakecasing routes;)

src/odm2_postgres_api/routes/loggers_routes.py
knl88: Any reason for using `time_series` and not `timeseries` as in the get?

src/odm2_postgres_api/metadata_init/data/loggers/loggers_metadata.py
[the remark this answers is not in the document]
haakondr: There are some options here:

- samplingfeatures.featuregeometry
- samplingfeatures.featuregeometrywkt
- sites (with long,lat) and spacialreference (coordinate system)

I am not sure about the motivation behind the sites table, I also think it is a bit of an overkill. Suggest we stick to using samplingfeatures table. Another thing is that sites table only allow for "point" type of locations, maybe we'd want to define a station as a wider area in the future?

Then, the question is, should we use featuregeometry column or featuregeometrywkt? These seem to both solve the same problem and a bit weird to have both. I seem to recall that WKT strings can embed information about coordinate system, while a postgis geometry column needs to specify the coordinate system (?). I do think that the geometry columns a better in that they allow for GIS based indexing and more ease of use querying. I am not a gis or postgis expect though so might be that we can do the same with WKT strings, but not sure.

Converting from lon/lat to wkt I suppose there are various libraries available. Seems like Roald used shapely for doing that, see https://github.com/NIVANorge/odm2-postgres-api/blob/dev/src/odm2_postgres_api/queries/core_queries.py#L385.

src/odm2_postgres_api/routes/loggers_routes.py
haakondr: agreed. Routes should use dash (-) as separator instead of underscore. It is not invalid to use underscores after the domain name in urls (I think), but common practice to use dash.

src/odm2_postgres_api/routes/loggers_routes.py
haakondr: FYI: I have a solution to building upsert queries in observations-api, but I am not 100% sure if it is sql-injection proof, needs to be looked at:

https://github.com/NIVANorge/observations-api/blob/fdf7d60897f9d5d57844f6e590f4181f3c867c1b/api/observations_api/utils/db_utils.py#L108. It is not trivial however, and we might want to either just write out the SQL ourselves or opt for an actual [ORM](https://en.wikipedia.org/wiki/Object%E2%80%93relational_mapping)

src/odm2_postgres_api/metadata_init/data/loggers/loggers_metadata.py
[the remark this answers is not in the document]
knl88: In [ODM2](https://www.sciencedirect.com/science/article/pii/S1364815216300093) table 1 there is some sample queries that I guess justifies the site table. Sites seems to be an old part of odm, and something I would from the name think is a good fit. But I need to understand the model better, btw, it is also mentioned in section 4.2.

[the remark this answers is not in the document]
knl88: Perhaps the idea is that you can have a samplingfeature with a site where the samplingfeature covers a wider region according to the feature geometry? and have several samplingfeatures for the same site

src/odm2_postgres_api/routes/loggers_routes.py
ZofiaRudjord: all routes ( shared routes , begroing, mass spec) are using "_" and not "-" in this api...I feel it is too late to start changing that.

src/odm2_postgres_api/metadata_init/data/loggers/loggers_metadata.py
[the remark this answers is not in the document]
knl88: For me feature geometry is fine for the MVP, and then we can see how to solve it with sites later😊Given that it doesn't give us too much extra work in the future.

PR #121, opened 2021-05-19
Adding get units end point which return all units

ZofiaRudjord:
A general question: is it better to create a new end point or extend an existing one to more functionalities ?
In this case there is an end-point /unit which takes two args: unittypecv and unitabbreviation, they would have to be made optional and the return type would have to be changed too.


src/odm2_postgres_api/queries/core_queries.py
```diff
@@ -50,6 +50,20 @@ async def find_unit(conn: asyncpg.connection, unit: UnitsCreate, raise_if_none=F
...
+    rows = await conn.fetch(f"SELECT * FROM units")
+
+    if rows is None and raise_if_none is True:
+        raise HTTPException(
+            status_code=422,
+            detail=f"No units registered in odm2",
+        )
+    for row in rows:
+        logging.info(f"fetching unit {row} ")
+        logging.info(f"Units(**row) {Units(**row)} ")
```
haakondr, line 63: may want to exclude logging

PR #122, opened 2021-05-28
Update to posting loggers data to odm2

ZofiaRudjord:
Removing end point for fetching ts id, instead it is a part of insert_timeseries query. Re-writting instert_timesries to accept a dict with variablecode and List of CreateTimeseriesResultValues in a request body. This allows for one hhtp request per logger file. Adding function to upsert time series. For now it does nothing on conflict so we can reingest without errors but details should be discussed.


src/odm2_postgres_api/queries/core_queries.py
```diff
@@ -30,6 +30,14 @@ def make_sql_query(table: str, data: dict):
     return f"INSERT INTO {table} ({','.join(data.keys())}) VALUES ({argument_placeholder(data)}) returning *"
 
 
+def upsert_sql_query(table: str, data: dict):
+    # https://www.postgresqltutorial.com/postgresql-upsert/
+    return (
+        f"INSERT INTO {table} ({','.join(data.keys())}) VALUES ({argument_placeholder(data)}) "
+        f"ON CONFLICT DO NOTHING returning *"
+    )
+
```
ZofiaRudjord, line 39: For now it does nothing on conflicts. I will prepare separate PR with modified version of this function but we should agree what the function should do on conflicts. I suggest it update qccodes, but perhaps it is safer to update entire row?

src/odm2_postgres_api/queries/loggers_queries.py
```diff
@@ -78,21 +80,26 @@ async def register_timeseries(
...
+                datavalue=item.datavalue,
+                valuedatetime=item.valuedatetime,
+                valuedatetimeutcoffset=item.valuedatetimeutcoffset,
+                censorcodecv=item.censorcodecv,
+                qualitycodecv=item.qualitycodecv,
+                timeaggregationinterval=item.timeaggregationinterval,
+                timeaggregationintervalunitsid=units.unitsid,
+            )
+            await core_queries.insert_timeseries_resultvalues(conn, timeseriesresultvaluecreate)
+    logging.info(f"Successfully ingested data for {platform_code}")
```
ZofiaRudjord, line 105: I am not totally sold on this implementation. But I discussed it briefly with @haakondr  this afternoon. Having insert inside python loop might be slow. On this other hand this implementation allows us to have the elegant type check and a well defined transation between this api and the nivacloud-loggers code. 

I suggest we monitor performance of this code and if we run into issue with speed we can revisit it.

ZofiaRudjord: This is the corresponding PR for the nivacloud-loggers: https://github.com/NIVANorge/nivacloud-loggers/pull/18

src/odm2_postgres_api/queries/core_queries.py
knl88: Initially I would think update entire row

knl88 (approved):
lgtm

Issue #120, opened 2021-05-11
timeseriesvalues should but upserted and not duplicated when QC values change.


ZofiaRudjord:
insert_timeseries_resultvalues will create a duplicate if data are ingested with updated QC value.
This should be changed to upsert.

ZofiaRudjord:
Currently there is the following unique constraint on the TimeSeriesResultValues: 
UNIQUE (ResultID, DataValue, ValueDateTime, ValueDateTimeUTCOffset, CensorCodeCV, QualityCodeCV, TimeAggregationInterval, TimeAggregationIntervalUnitsID));

This means that if we update the QC flag ( eg due to changing thresholds ) we will end up having 
for the same timeseries type ( e.g. temp) for a given timestamp, value pair two rows: with good and with bad qc.
This makes no sense. 

Do you know if this constraint was created by Roald or automatically generated @haakondr ?
If later what is the purpose of it?

I suggest to alter it to:
UNIQUE (ResultID, DataValue, ValueDateTime, ValueDateTimeUTCOffset, CensorCodeCV, TimeAggregationInterval, TimeAggregationIntervalUnitsID));
and update QualityCodeCV when re-running QC.
Before I proceed with modyfying this unique constraints I would like to hear your opinion on this @haakondr @knl88 @jemmima

haakondr:
Agree to changing the unique constraint to how you propose. Does not make sense to include QualityCodeCV in the constraint.

according to git history this comes from when Roald copied over the odm2 schema "as is".

PR #123, opened 2021-06-02
Upserting QC

ZofiaRudjord:
Adding a new unique constraint on TimeSeriesResultValues which required all column values except qualitycodecv to be unique, updating qc on violation of this constraint. Solves #120


src/odm2_postgres_api/db_init/ODM2_for_PostgreSQL.sql
```diff
@@ -2368,3 +2368,5 @@ foreign key (RelationshipTypeCV) References ODM2.CV_RelationshipType (Name)
 on update no Action on delete RESTRICT;
 
 alter table ODM2.Directives add column directivename varchar (255) UNIQUE;
+
+ALTER TABLE ODM2.TimeSeriesResultValues ADD CONSTRAINT all_but_qc_unique UNIQUE(ResultID, DataValue, ValueDateTime, ValueDateTimeUTCOffset, CensorCodeCV, TimeAggregationInterval, TimeAggregationIntervalUnitsID);
```
knl88, line 2372: Don't think all_but_qc_unique follows the naming for used in the schema. Should probably be something like fk_TimeSeriesResultValues_....

haakondr: fk_ indicates that it is a foreign key, and in this case it is not, it is a unique constraint.

Quick googling suggests the following naming convention:

UQ_<TableName>_<ColumnName(s)>

I don't have any strong opinions on this, but it is perhaps a good idea to continue on the naming convention used (<type of constraint>_<table>_<columns>

haakondr: Kim pointed out that this does not drop the current existing unique constraint, but only creates a new one.

Will need to add a statement which drops timeseriesresultvalues_resultid_datavalue_valuedatetime_val_key here. Since this script runs on each app startup, it would have to also work if it does not exist.

haakondr, line 2372: May want to omit CensorCodeCV from this unique constraint? I am not 100% sure about the usage of censorcode, but to me it seems like we'd only want one value per resultid/time.

src/odm2_postgres_api/queries/core_queries.py
```diff
@@ -498,7 +499,7 @@ async def insert_timeseries_resultvalues(
 ):
     async with conn.transaction():
         data = {k: v for k, v in timeseries_resultvalues}
-        return await conn.fetchrow(upsert_sql_query("timeseriesresultvalues", data), *data.values())
+        return await conn.fetchrow(upsert_ts_all_but_qc_unique("timeseriesresultvalues", data), *data.values())
```
haakondr, line 502: as discussed with kim, would be good to have a unit test on this function. should be fairly straight forward, can copy the setup in test_storage.py or any of the other files. 

Running tests is not really documented, there are some quirks, I went through most of them with Kim. 

- working directory needs to be in the project root (so that integration-test-docker-compose.yml gets properly picked up).
- a new test file needs to import the following for the fixtures to work:

```
from integration_test_fixtures import wait_for_db, db_conn
```

This would be solved by renaming integration_test_fixtures.py -> conftest.py

- pytest need to be executed with  `-m docker` in order to run the tests annotated with @pytest.mark.docker

haakondr (changes requested):

src/odm2_postgres_api/queries/core_queries.py
```diff
@@ -30,11 +30,12 @@ def make_sql_query(table: str, data: dict):
...
 
 
-def upsert_sql_query(table: str, data: dict):
+def upsert_ts_all_but_qc_unique(table: str, data: dict):
     # https://www.postgresqltutorial.com/postgresql-upsert/
     return (
         f"INSERT INTO {table} ({','.join(data.keys())}) VALUES ({argument_placeholder(data)}) "
-        f"ON CONFLICT DO NOTHING returning *"
+        f"ON CONFLICT ON CONSTRAINT all_but_qc_unique DO UPDATE SET "
+        f"qualitycodecv='{data['qualitycodecv']}' returning *"
```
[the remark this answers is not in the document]
knl88: One option could also be to use excluded.qualitycodecv,  created a branch with it https://github.com/NIVANorge/odm2-postgres-api/blob/7d8660e26a54452935d94dbeb325b77b2d0d2811/src/odm2_postgres_api/queries/core_queries.py#L40

src/odm2_postgres_api/db_init/ODM2_for_PostgreSQL.sql
ZofiaRudjord: I had a problem to drop the constraint, probably syntax again...
In any case I left the constraint because it is less restrictive than the one I added. So my logic was that the one I added will fire first no matter and I tested the code that I can upsert. 
Do you think it is important to delete the constraint?

ZofiaRudjord: Regarding naming, it is a good idea to have a system for that. Prepending with uq_ sounds good to me.
How about uq_all_but_qc ?

haakondr: about not dropping constraint: we could probably keep it, don't think it matters much

src/odm2_postgres_api/queries/core_queries.py
knl88: @ZofiaRudjord  I Added some tests to https://github.com/NIVANorge/odm2-postgres-api/blob/7d8660e26a54452935d94dbeb325b77b2d0d2811/tests/test_core_queries.py#L234 if you want them:)

src/odm2_postgres_api/queries/core_queries.py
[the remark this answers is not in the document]
haakondr: hm yes, that approach could probably work. I have not used EXCLUDED before, but seems useful.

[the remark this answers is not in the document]
ZofiaRudjord: I updated the code to use the EXCLUDED alias as suggested by @knl88 . I updated however, his proposed unique constraint to be applied to all columns but the two qc ones:  qualitycodecv and CensorCodeCV and therefore to update only these two values on conflict.

src/odm2_postgres_api/queries/core_queries.py
ZofiaRudjord: I merged @knl88 tests to this branch.

src/odm2_postgres_api/db_init/ODM2_for_PostgreSQL.sql
ZofiaRudjord: Agree, this code is now updated as well and excluded from unique constraint.

tests/test_core_queries.py
```diff
@@ -191,3 +192,53 @@ def test_create_taxons_with_wrong_annotation():
...
+async def test_upsert_timeseries_resultvalues(db_conn, timeseries_base_data, new_key, new_value, expected):
+    base_row = await upsert_tsrv_table(db_conn, timeseries_base_data)
+    timeseries_base_data[new_key] = new_value
+    actual_row = await upsert_tsrv_table(db_conn, timeseries_base_data)
+
+    assert actual_row[new_key] == expected
+    assert base_row["valueid"] == actual_row["valueid"]
+
+    for k in base_row.keys():
+        if k != new_key:
```
haakondr, line 244: so, the behaviour is that datavalue is never updated, but silently ignored in case of duplicates (?)

Suggest we look at a way to throw errors if it happens that a) we already have a value  and b) the value differs. But, I don't think this blocks a merge and is only a future improvement.

tests/test_core_queries.py
```diff
@@ -191,3 +192,53 @@ def test_create_taxons_with_wrong_annotation():
...
+            assert base_row[k] == actual_row[k]
```
haakondr, line 245: Looks good. 

Possible future improvements on the tests: Could be good to expand this test to check that the values are actually stored. Here we are only testing the response.

Could the core_queries.find_timeseries function be suitable for this? That would also add test coverage to this function.

haakondr (approved):

tests/test_core_queries.py
ZofiaRudjord: The idea is that it never happens , but you are right, the code should throw error if/when it happens. :-)

knl88 (approved):
lgtm, guess I can try to wrap my head around other test cases when needed:)

tests/test_core_queries.py
knl88: Yeap, guess this can give various unintended effects that we need to think of in the future.

PR #146, opened 2021-06-28
Adding fetch data route for loggers data

ZofiaRudjord:


src/odm2_postgres_api/routes/loggers_routes.py
```diff
@@ -12,3 +12,10 @@ async def post_timeseries_result_values(
     data: Dict[str, List[schemas.LoggersTimeSeriesResultValuesCreate]], connection=Depends(api_pool_manager.get_conn)
 ):
     await insert_timeseries(connection, data)
+
+
+@router.post("/fetch_timeseries_result_values")
```
haakondr, line 17: if we can use GET requests for queries, that would be very good as they add all the queries within the url as query/path parameters. The benefit of that is that the URL can be easily shared to others. 

I think in this case that it can be doable to model this as a GET request. This is more of a preference thing though, not very important.

haakondr, line 17: Like previously discussed with Kim, I think url paths should use dash instead of underscore as this is more common. 

I think there are two reasons behind this:

1. underscores are not allowed in domains, so you could end up with something like https://my-domain.com/some_path. Not end of the world, but good to be consistent
2. links on the web are typically underlined making the underscore indistinguishable to a space

Probably many other reasons, but I guess the most important one is to be consistent

haakondr (approved):

src/odm2_postgres_api/routes/loggers_routes.py
ZofiaRudjord: The reason why I have used "_" instead of "-" is that all routes ( started by those introduced by Roald in the shared routes) use "_" so I wanted to stay consistent with that. Perhaps we can redo all routes on a separate branch @haakondr  ?

haakondr:
Looking at your updated code, you should not be getting duplicates (unless there really are). Could be something wrong with your query

ZofiaRudjord:
Results is a concat of several tables which are used in join, resultid is the one which is duplicated in result, timeseriesresult and timeseriesresultvalues. 
If you think that this is wrong that the result of fetch is the concat of all these tables leading to duplicates please write me back @haakondr . I might be missing sth.

haakondr:
> Results is a concat of several tables which are used in join, resultid is the one which is duplicated in result, timeseriesresult and timeseriesresultvalues.
> If you think that this is wrong that the result of fetch is the concat of all these tables leading to duplicates please write me back @haakondr . I might be missing sth.

I think that this can be solved by using the right set of left/right joins in the query. Alternatively, picking another table to _query from_ as in the starting point of the query.

A bit difficult to point to what to change without having a live database with data to query against, as my sql foo is not that advanced. I can probably test against the nivatest instance?

PR #147, opened 2021-06-29
Feature/add qc tests

ZofiaRudjord:


src/odm2_postgres_api/metadata_init/data/loggers/loggers_metadata.py
```diff
@@ -557,6 +576,7 @@ def loggers_timeseries_result() -> List[LoggersTimeSeriesCreate]:
             "unitstypecv": "Length",
             "unitsabbreviation": "mm",
             "aggregationstatisticcv": "Average",
+            "dataqualitycodes": ["frozen_test", "snowdepth_variability_test"],
         },
     ]
-    return [LoggersTimeSeriesCreate(**ts) for ts in timeseries]
+    return [LoggersTimeSeriesCreate.parse_obj(ts) for ts in timeseries]
```
ZofiaRudjord, line 582: parse_obj was suggested by @knl88  to solve and issue that an error message: 
 error: Argument after ** must be a mapping, not "object"
was thrown by mypy on 582 when trying to map ts to LoggersTimeSeriesCreate

src/odm2_postgres_api/routes/loggers_routes.py
```diff
@@ -19,3 +23,8 @@ async def fetchtimeseriesresultvalues(
     data: schemas.LoggersTimeSeriesResultGetValues, connection=Depends(api_pool_manager.get_conn)
 ):
     return await get_timeseriesresultvalues(connection, data)
+
+
+@router.post("/fetch_timeseries_qc_tests")
```
ZofiaRudjord, line 28: all routes in odm2 use underscore not a hyphen. If disturbed by this naming I suggest to open an issue and change all of them in one go.

src/odm2_postgres_api/routes/loggers_routes.py
```diff
@@ -19,3 +23,8 @@ async def fetchtimeseriesresultvalues(
...
+async def fetchtimeseriesqctests(data: schemas.LoggersTimeSeriesCreate, connection=Depends(api_pool_manager.get_conn)):
```
haakondr, line 29: May introduce a specific class for this query?

From what I can tell, only three fields are used in the actual query:

- samplingfeaturecode
- variablecode
- processinglevelcode

LoggersTimeSeriesCreate has a lot of fields not used in the endpoint and it seems unnecessary to have to supply them. I do note that the other fields default to other values, but I think it would be a good idea to introduce a specific type for this.

haakondr: May perhaps also just change it to a GET request with samplingfeaturecode, variablecode and processinglevelcode as string params?

haakondr (approved):
I am not 100% up to date about the usage of this endpoint, but I am guessing this is the equivalent of the "get list of QC tests from metaflow" in the ferrybox pipeline. In other words, the API dictates which QC tests to run on a given measurement.

I do think that the approach in ferrybox pipeline is a bit "chatty" in that the signals2tsb deployment queries for the same information in metaflow many, many times. I have not really thought out a solution, other than that our approach was not ideal. 

What do you think? I don't think we need to resolve this now, but I imagine a solution with a solid cache that invalidates on changes in a good way. 

One solution could be to setup a github action (cronjob) that queries this API endpoint on interval, and on changes commits a json list of QC tests in our data pipeline, thus triggering a redeploy of our pipeline.  This way, metadata is hardcoded in code, but updated on changes. Good fit for _nearly static_ data.

src/odm2_postgres_api/queries/loggers_queries.py
```diff
@@ -133,3 +134,20 @@ def remove_duplicates(my_dict):
         return unique_dict
 
     return [schemas.TimeSeriesResultValues(**remove_duplicates(result)) for result in results] if results else []
+
+
+async def get_timeseries_qc_tests(conn: asyncpg.connection, ts: schemas.LoggersTimeSeriesCreate) -> List[str]:
+    results = await conn.fetch(
+        "select dataqualitycode from dataquality dq "
```
haakondr, line 141: may want to include a bit more than just the dataqualitycode? I interpret this as "get relevant metadata for a given timeseries"? May also want to move this to "shared_routes" as there is nothing logger specific about this endpoint that I can see (?)

src/odm2_postgres_api/metadata_init/data/loggers/loggers_metadata.py
ZofiaRudjord: The problem I ran into was that pydantic dynamic casting did not recognise List as a valid value in the dict so I got an error Argument after ** must be a mapping, not "object" . parse_obj solves that (in addition to adding useful error messages :) )

knl88: Yes that should be correct. I this case it is used to avoid the mypy error when unpacking, not completely sure, but guess it only expands the top-level key-value pairs for type checking and throws an error on the list.

src/odm2_postgres_api/metadata_init/data/loggers/loggers_metadata.py
```diff
@@ -36,6 +37,24 @@ def loggers_sampling_features() -> List[SamplingFeaturesCreate]:
...
+            "dataqualitytypecv": "Accuracy",
+            "dataqualitycode": "frozen_test",
+            "dataqualitydescription": "Five (or more) identical consecutive measurements indicates "
+            "that sensor is mal-functioning (frozen).",
+        },
+        {
+            "dataqualitytypecv": "Accuracy",
+            "dataqualitycode": "snowdepth_variability_test",
+            "dataqualitydescription": "Measurements of snow depth which show high variability of depth gradient"
+            "over time are inaccurate.",
```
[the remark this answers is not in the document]
ZofiaRudjord: I agree with keeping it specific and if need be make it more general.

knl88 (approved):
lgtm:)

src/odm2_postgres_api/queries/loggers_queries.py
ZofiaRudjord: Hm...do you have a specific info in mind? I thought of this end point as intended to provide a list of data quality tests
to run. I could return all content of data quality table, is that what you have in mind?

Good point to move it to shared_routes, it can be used for ctd, ferryboxes, and other..

ZofiaRudjord:
> I am not 100% up to date about the usage of this endpoint, but I am guessing this is the equivalent of the "get list of QC tests from metaflow" in the ferrybox pipeline. In other words, the API dictates which QC tests to run on a given measurement.
> 
> I do think that the approach in ferrybox pipeline is a bit "chatty" in that the signals2tsb deployment queries for the same information in metaflow many, many times. I have not really thought out a solution, other than that our approach was not ideal.
> 
> What do you think? I don't think we need to resolve this now, but I imagine a solution with a solid cache that invalidates on changes in a good way.
> 
> One solution could be to setup a github action (cronjob) that queries this API endpoint on interval, and on changes commits a json list of QC tests in our data pipeline, thus triggering a redeploy of our pipeline. This way, metadata is hardcoded in code, but updated on changes. Good fit for _nearly static_ data.

I have not thought of it but it is certainly a good idea to cache it and update the cash when update to qc tests appears.
Just FYI, "chattyness" will be (for now) less of an issue for loggers, because the data transfer frequency is much lower.

PR #157, opened 2021-08-19
dev to master

ZofiaRudjord:


src/odm2_postgres_api/metadata_init/data/loggers/loggers_metadata.py
```diff
@@ -332,6 +332,13 @@ def loggers_variables() -> List[VariablesCreate]:
             "variablecode": "TempCond_Avg",
             "nodatavalue": -9999,
         },
+        {
+            "variabletypecv": "Chemistry",
+            "variablenamecv": "Temperature, sensor",
+            "variabledefinition": "Temperature measured by Ph sensor",
```
haakondr, line 338: question: does it make sense to differentiate temperature variables by how they are measured? Does temp measured by a ph sensor differ from another sensor? I may not be seeing the full picture here

haakondr (approved):
lgtm

src/odm2_postgres_api/metadata_init/data/loggers/loggers_metadata.py
ZofiaRudjord: It is important for operation people (like Uta). I opened an issue about it in nivacloud-loggers: https://github.com/NIVANorge/nivacloud-loggers/issues/39

ZofiaRudjord: But this is where cv' come in handy. We can always query all time series's which measure temperature by
putting a constraint on "variablenamecv"

PR #161, opened 2021-09-02
Adding onumbers column to directives table to store a list of O-numbers

ZofiaRudjord:
This column is necessary to store O numbers (stored in nivabase projects table).


jemmima (approved):
I tested running db_init and also testing posting to directives endpoint to add a new project and it all worked ok 👍

PR #162, opened 2021-09-08
Adding end point to fetch all time series emasured at a given site

ZofiaRudjord:
This end point is needed for qc module of loggers.


src/odm2_postgres_api/queries/loggers_queries.py
```diff
@@ -134,3 +134,16 @@ def remove_duplicates(my_dict):
...
+    results = await conn.fetch(
+        "select variablecode from variables v "
+        "join results r on r.variableid = v.variableid "
+        "join timeseriesresults tsr on r.resultid = tsr.resultid "
+        "join featureactions f on f.featureactionid = r.featureactionid "
+        "join samplingfeatures sf on f.samplingfeatureid=sf.samplingfeatureid "
+        "where sf.samplingfeaturecode=$1 ",
+        platformcode,
+    )
+    return [result["variablecode"] for result in results] if results else []
```
haakondr, line 149: I think the else clause here is redundant, I think this should be the same:

```
return [result["variablecode"] for result in results]
```

haakondr (approved):

src/odm2_postgres_api/routes/loggers_routes.py
```diff
@@ -22,3 +23,8 @@ async def fetchtimeseriesresultvalues(
     data: schemas.LoggersTimeSeriesResultGetValues, connection=Depends(api_pool_manager.get_conn)
 ):
     return await get_timeseriesresultvalues(connection, data)
+
+
+@router.get("/get_timeseries_list/{platformcode}")
```
haakondr, line 28: should perhaps name this endpoint something like `/variables/{samplingfeaturecode}`? I guess this is a question if we want application specific routes or generic routes in our API. I think I prefer more generic routes as odm2 is quite generic, but either works

ZofiaRudjord: Yes, but this is not all variables but those which result in time series ( in contrary to depth profiles or lab measurements). Do you still think it will not be confusing to keep it generic (since it is not all variables) ?

haakondr: hm, that's right. So, only variables for a samplingfeature that do have timeseriesresults

src/odm2_postgres_api/queries/loggers_queries.py
```diff
@@ -134,3 +134,16 @@ def remove_duplicates(my_dict):
         return unique_dict
 
     return [schemas.TimeSeriesResultValues(**remove_duplicates(result)) for result in results] if results else []
+
+
+async def get_timeserieslist(conn: asyncpg.connection, platformcode: str) -> List[str]:
+    results = await conn.fetch(
+        "select variablecode from variables v "
+        "join results r on r.variableid = v.variableid "
+        "join timeseriesresults tsr on r.resultid = tsr.resultid "
```
haakondr, line 143: since we join on timeseriesresults, this will only work once we have values. Does that cause any problems?

ZofiaRudjord: The list will be empty if there is no timeseries results, that's the intension.

PR #168, opened 2021-09-29
Feature/kafka consumer

jemmima:
@ZofiaRudjord @haakondr, we have already added the `kafka_records` table to the odm2 schema for storing the raw kafka messages, and I'm also adding one to store partial stations until the rest arrives before inserting into `samplingfeatures` table. These two tables are very `kafka-sink` specific, and they don't belong in the odm2 schema. Should we make a separate schema for this?


haakondr:
I suppose we can have a separate schema, that can make sense.

What is the need for this temporary partial stations table, is the kafka_records table not sufficient?

jemmima:
> I suppose we can have a separate schema, that can make sense.
> 
> What is the need for this temporary partial stations table, is the kafka_records table not sufficient?

I wanted to ask you about it - will kafka_records just persist all the messages indefinitely? I couldn't find any code anywhere that removes rows, but I imagine we will want to add that at some point? I thought an extra table for partial stations could hold only unmatched station or project_station messages until the other arrives, and then those rows can be removed when their match comes and they are added to the samplingfeatures table. This was my thinking but I'm eager for feedback on it! :)

haakondr:
> > I suppose we can have a separate schema, that can make sense.
> > What is the need for this temporary partial stations table, is the kafka_records table not sufficient?
> 
> I wanted to ask you about it - will `kafka_records` just persist all the messages indefinitely? I couldn't find any code anywhere that removes rows, but I imagine we will want to add that at some point? I thought an extra table for partial stations could hold only unmatched `station` or `project_station` messages until the other arrives, and then those rows can be removed when their match comes and they are added to the `samplingfeatures` table. This is because we don't want partial stations in samplingfeatures table. This was my thinking but I'm eager for feedback on it! :)

I was thinking that we just store the kafka records forever. Depends on the volume, but I'd think we're talking about megabytes. In other words, small volumes.

jemmima:
> > > I suppose we can have a separate schema, that can make sense.
> > > What is the need for this temporary partial stations table, is the kafka_records table not sufficient?
> > 
> > 
> > I wanted to ask you about it - will `kafka_records` just persist all the messages indefinitely? I couldn't find any code anywhere that removes rows, but I imagine we will want to add that at some point? I thought an extra table for partial stations could hold only unmatched `station` or `project_station` messages until the other arrives, and then those rows can be removed when their match comes and they are added to the `samplingfeatures` table. This is because we don't want partial stations in samplingfeatures table. This was my thinking but I'm eager for feedback on it! :)
> 
> I was thinking that we just store the kafka records forever. Depends on the volume, but I'd think we're talking about megabytes. In other words, small volumes.

Ah ok! I'll just use that then - is there any performance issue searching the unindexed payload column for matching `id`s embedded in the json if the table gets very large?

haakondr:
Can add index on json fields in postgres. That is probably a good idea

haakondr:
@jemmima I assume we can close this as this work got continued on the nivabase-sync branch (?)

PR #169, opened 2021-10-04
Updating query get_timeseriesresultvalues to accept also n_historical instead of starttime

ZofiaRudjord:
This update is needed when fetching data for qc tests which require certain number of historical samples to run (e.g. frozen test or spike test).


ZofiaRudjord:
I will merge this to for further testing but comments are welcome even if this is merged.

PR #170, opened 2021-10-05
mv integration-test compose file

knl88:
Getting issues with the location of the integration-test-docker-compose.yml file. Quickfix is to move the file to the root of where pytest run.

PR #171, opened 2021-10-06
Pytest docker compose fix

knl88:
This is a fairly strange and confusing problem, but the goal is to be able to run the integration tests. That should be possible to do with `pytest -m docker` both locally and on CircleCI. 

Initially in PR #170 I just moved the integration-test-docker-compose.yml file to the root directory. That makes it possible to run the integration tests, but it picks up the docker-compose.yml file and the timescale_odm2 container from there. It also fails on the first attempt to run the tests on circleci.

This PR is a workaround for that problem by renaming tests/integration-test-docker-compose.yml -> docker-compose.yml and giving the test container the name timescale_odm2_pytest, to avoid confusion about what container is running.

For me real problem seems to be related to the passing of the docker-compose file to pytest-docker-compose plugin. As the plugin seems to be very fond of files named docker-compose.yml in the root dir:) Perhaps it is better to skip the compose plugin, and use the pytest-docker plugin for the timescale_odm2 container.


ZofiaRudjord (approved):
Thanks for the fix.
Ideally I would prefer to keep docker-compose.yml name for local build to avoid the hassle of passing file arg (-f odm2-docker-compose.yml). 
But in order to be able to move on, let's merge this fix now:-)

PR #172, opened 2021-10-08
dev to master (odm2)

ZofiaRudjord:


odm2-docker-compose.yml
```diff
@@ -0,0 +1,41 @@
+version: '3.5'
```
haakondr, line 1: does this mean that we have to specify `-f odm2-docker-compose.yml` when running normally?

If I have understood the issue, is that running pytest in subdirectory fails as the path to docker-compose is wrong (relative to current working directory). With this change, we get a similar problem when attempting to run regular docker-compose. 

I think we should prioritize normal running over test running. Afterall, running tests do work on CI (from the root directory). I probably don't have the full context here @knl88  and @ZofiaRudjord, so please fill me in.

src/odm2_postgres_api/db_init/ODM2_for_PostgreSQL.sql
```diff
@@ -2369,4 +2369,5 @@ on update no Action on delete RESTRICT;
 
 alter table ODM2.Directives add column directivename varchar (255) UNIQUE;
 
-alter table ODM2.TimeSeriesResultValues add constraint uq_TimeSeriesResultValues_all_but_qc UNIQUE(ResultID, ValueDateTime, ValueDateTimeUTCOffset);
\ No newline at end of file
+alter table ODM2.TimeSeriesResultValues add constraint uq_TimeSeriesResultValues_all_but_qc UNIQUE(ResultID, ValueDateTime, ValueDateTimeUTCOffset);
+alter table ODM2.Directives add column onumbers varchar (255);
```
haakondr, line 2373: what is the convention on onumbers, commaseparated?

Ideally, I think we represent lists as a structured list somehow. I have zero experience with [postgres array types](https://www.postgresql.org/docs/9.1/arrays.html) but they could maybe work?

ZofiaRudjord: Good point. I suggest we revert altering Directives. @jemmima is working on kafka consumer which needs to store onummbers. Perhpas she has a different/better solution.

odm2-docker-compose.yml
haakondr: Last successful build seems to be https://app.circleci.com/pipelines/github/NIVANorge/odm2-postgres-api/992/workflows/1309facd-205a-48f5-82e5-0a47c2ebcb78. (without @knl88  changes) 

I tried rerunning this revision and that now failed: https://app.circleci.com/pipelines/github/NIVANorge/odm2-postgres-api/992/workflows/b7b78994-593a-4bfb-9b49-a59fcd0572fb

Not sure what has happened here, I will have a look.

src/odm2_postgres_api/queries/core_queries.py
```diff
@@ -60,6 +62,19 @@ async def find_unit(conn: asyncpg.connection, unit: UnitsCreate, raise_if_none=F
...
+    row = await conn.fetchrow(
+        f"SELECT * FROM resultsdataquality WHERE resultid=$1 AND dataqualityid=$2", rdq.resultid, rdq.dataqualityid
+    )
+    if row is None and raise_if_none:
+        raise HTTPException(
+            status_code=422,
+            detail=f"resultsdataquality resultid={rdq.resultid} and "
+            f"dataqualityid={rdq.dataqualityid} does not exist",
+        )
+    return ResultsDataQuality(**row) if row else None
```
jemmima, line 75: Is it worth using results_data_quality_create instead of rdq for readability?

jemmima (approved):
LGTM!

src/odm2_postgres_api/db_init/ODM2_for_PostgreSQL.sql
ZofiaRudjord: Ok, so I keep it as is, to be changed when kafka consumer is ready

PR #176, opened 2021-10-28
Allow longer samplingfeaturecode and add combinefeatures method.

jemmima:
Linked prs:
https://github.com/NIVANorge/nivacloud-mass-spectrometry/pull/40
https://github.com/NIVANorge/mass-spec-ui/pull/80


src/odm2_postgres_api/db_init/ODM2_for_PostgreSQL.sql
```diff
@@ -209,7 +209,7 @@ create table ODM2.SamplingFeatures (
 	samplingfeatureid serial  NOT NULL primary key,
 	samplingfeatureuuid uuid  NOT NULL,
 	samplingfeaturetypecv varchar (255) NOT NULL,
-	samplingfeaturecode varchar (50) NOT NULL,
+	samplingfeaturecode varchar (100) NOT NULL,
 	samplingfeaturename varchar (255) NULL,
```
jemmima, line 213: this is to allow for some of the longer filenames from mass spec - we should think about changing this later but for now the whole pipeline is set up to have the filename stored as the samplingfeaturecode.

ZofiaRudjord (approved):
LGTM!

src/odm2_postgres_api/db_init/ODM2_for_PostgreSQL.sql
haakondr: This won't be applied on an existing database. This script just run plain sql statements and fail (gracefully) if table exist.

In other words, if ODM2.SamplingFeatures table already exists you need to perform a alter table statement, something like:

```
alter table ODM2.SamplingFeatures alter column samplingfeaturecode type varchar (100)
```

I have not tried this. need to ensure that not null constraint is not overwritten by this.

haakondr (changes requested):
the database change won't be applied, see my other comment

src/odm2_postgres_api/db_init/ODM2_for_PostgreSQL.sql
haakondr: I think longer codes is ok for now, but should probably aim to keep them below 50 or so, as UI's displaying codes typically don't reserve a lot of space to display codes as they should ideally be quite short.

ZofiaRudjord:
Thanks for catching this @haakondr !  the reason it works locally is because with doco down we wipe the db.

src/odm2_postgres_api/db_init/ODM2_for_PostgreSQL.sql
jemmima: @ZofiaRudjord is it better to restrict the length of the filename to 50 chars do you think?

ZofiaRudjord: Ideally yes, but you had a reason for needing longer codes, right?

jemmima: Yes only that some of the test file names had too long codes, so I anticipated this might happen a lot, but might be better to restrict it?

jemmima: Implemented in 113eabc. I tested this locally on an existing odm2 instance and the column is still UNIQUE after, so I'm going to merge it and test in niva test.

PR #185, opened 2021-12-20
dev to master

ZofiaRudjord:
I would like to merge asap to connect the remaining logger (Kraakfoss).
If you want more time to review I will pick just the last commit to merge to master.


src/odm2_postgres_api/metadata_init/data/cdom/cdom_metadata.py
```diff
@@ -0,0 +1,36 @@
...
+    ]
+    return [SamplingFeaturesCreate(**sf) for sf in sampling_features]
+
+
+def cdom_directives() -> List[DirectivesCreate]:
+    cdom_directives_list = [
+        {
+            "directivetypecv": "Project",
+            "directivename": "Vaarflom",
+            "directivedescription": "Should be merged with the revenat project in nivabase after sink",
```
haakondr, line 28: could perhaps be good to add the project_id in this comment to ease fixup later

src/odm2_postgres_api/queries/core_queries.py
```diff
@@ -42,6 +45,14 @@ def upsert_timeseries_values(table: str, data: dict):
     )
 
 
+def upsert_spectra_values(table: str, data: dict):
+    # https://www.postgresqltutorial.com/postgresql-upsert/
+    return (
+        f"INSERT INTO {table} ({','.join(data.keys())}) VALUES ({argument_placeholder(data)}) "
+        f"ON CONFLICT DO NOTHING"
+    )
```
haakondr, line 53: be aware: i think we should be careful about doing this kind of string interpolation when building queries. If not done right, we risk SQL injection attacks. From what I can tell here, it should not be possible for the client to alter any of the params as they are either hardcoded or typechecked in API, but still somewhat risky.

haakondr (approved):

PR #188, opened 2022-01-07
moved odm2 API to a subdirectory "api"

haakondr:


.circleci/deployment_config.json
```diff
@@ -2,9 +2,9 @@
...
     "odm2api": {
       "image_name": "ODM2_API_IMAGENAME",
-      "docker_file_directory": ".",
+      "docker_file_directory": "api/",
       "kubectl_deployment_directories": "k8s/deployments/",
-      "is_changed_regexp": "."
+      "is_changed_regexp": "^api/|^k8s/"
     }
   }
 }
```
haakondr, line 10: this deployment config is not properly documented as far as I know, but in short it instructs circleci on how and when different jobs should start.

haakondr:
I see that the circleci task [run_pytest_docker](https://github.com/NIVANorge/niva-circleci-orbs/blob/f226abb4842fdc649a59251f7c94b6af56f189d2/src/niva-google-cloud-orb.yml#L44) expects files to be in the repo root. I'll see if I can find an easy fix, but these scripts are not very easy to reason about

Issue #195, opened 2022-01-31
kafka-sink: CardinalityViolationError: more than one row returned by a subquery used as an expression


haakondr:
```
message: "Failed to insert row, ignoring. Row: <Record topic='nivadatabase.project-stations' key='58945' payload='{"time": "2022-01-30T21:01:54.445889", "triggeredBy": "niva-data-producer", "action": "CREATE", "data": {"id": 58945, "projectId": 12510, "stationId": 73564, "stationCode": "SSVLITST1", "stationName": " S\\u00e6ters\\u00e6tervatnet litoral st.1", "enteredDate": "2022-01-30", "auditId": null}}'>"
```

```
Traceback (most recent call last):
  File "/app/kafka_sink/message_handlers.py", line 307, in station_post_processing
    await save_project_station(conn, msg)
  File "/app/kafka_sink/message_handlers.py", line 361, in save_project_station
    samplingfeature = await connection.fetchrow(
  File "/usr/local/lib/python3.9/site-packages/asyncpg/connection.py", line 641, in fetchrow
    data = await self._execute(
  File "/usr/local/lib/python3.9/site-packages/asyncpg/connection.py", line 1625, in _execute
    result, _ = await self.__execute(
  File "/usr/local/lib/python3.9/site-packages/asyncpg/connection.py", line 1650, in __execute
    return await self._do_execute(
  File "/usr/local/lib/python3.9/site-packages/asyncpg/connection.py", line 1697, in _do_execute
    result = await executor(stmt, None)
  File "asyncpg/protocol/protocol.pyx", line 201, in bind_execute
asyncpg.exceptions.CardinalityViolationError: more than one row returned by a subquery used as an expression
```

see https://console.cloud.google.com/logs/query;query=resource.labels.container_name%3D%22odm2-kafka-sink%22%0Aseverity%3DERROR%0AjsonPayload.exc_info:%22CardinalityViolationError:%20more%20than%20one%20row%20returned%20by%20a%20subquery%20used%20as%20an%20expression%22;timeRange=P2D;cursorTimestamp=2022-01-30T23:04:37.462781501Z?project=nivatest-1

haakondr:
I think this was caused by previous versions of the code producing wrong results, and then fixed by deleting nivatest projectstations.

However, I think we need a unique constraint here:

```
unique(projectstationid, externalidentifiersystemid)
```

We should never have more than one nivabase id for a given project station.

[...]

PR #199, opened 2022-02-01
added unique constraints on externalidentifers #195

haakondr:
projects, stations and projectstations should only have one external
identifier in a given system. Could be some edge cases here, but I think
it makes sense to start out making this unique.


jemmima (approved):
LGTM!

PR #204, opened 2022-02-02
203 deletes

haakondr:
changes:

- Handle deletes
  - somewhat primitive implementation, may want to generalize this a bit but thought I'd keep it simple
- perform post-processing in batch
I noticed that things were going quite slow. Changed the post-processing so that it queries kafka_record once per 1000 msg and not 1 per message. 

This speed up things quite a bit. Takes about 5 mins to ingest all data now (locally). Not sure how slow it was before this change, but something like 25 mins or so which was a bit slow


haakondr:
@jemmima i was thinking to merge this to dev so that I can fully ingest and verify written data but making PR so that you can look through and keep track of changes

PR #207, opened 2022-02-07
reverted back in cdom code which was accidentally removed

haakondr:
reverted back code from revision
5f25dcfe9f97465f1641227b8cdbed2543255db5 related to cdom. The code was
removed in commit
https://github.com/NIVANorge/odm2-postgres-api/commit/925941d731627b290e47853aef735398f118d1ea
so I took the code from the revision before this one


ZofiaRudjord (approved):
I suggest we merge and test on test cluster.

Issue #201, opened 2022-02-02
multiple samplingfeatures for one nivabase station_id cause errors


haakondr:
we get the following stacktrace:

```
Traceback (most recent call last):
  File "/app/kafka_sink/message_handlers.py", line 333, in project_post_processing
    await save_project_station(conn, msg)
  File "/app/kafka_sink/message_handlers.py", line 364, in save_project_station
    samplingfeature = await connection.fetchrow(
  File "/usr/local/lib/python3.9/site-packages/asyncpg/connection.py", line 679, in fetchrow
    data = await self._execute(
  File "/usr/local/lib/python3.9/site-packages/asyncpg/connection.py", line 1659, in _execute
    result, _ = await self.__execute(
  File "/usr/local/lib/python3.9/site-packages/asyncpg/connection.py", line 1684, in __execute
    return await self._do_execute(
  File "/usr/local/lib/python3.9/site-packages/asyncpg/connection.py", line 1731, in _do_execute
    result = await executor(stmt, None)
  File "asyncpg/protocol/protocol.pyx", line 201, in bind_execute
asyncpg.exceptions.CardinalityViolationError: more than one row returned by a subquery used as an expression
```

There are some stations which map to multiple samplingfeatures:

```
select samplingfeatureexternalidentifier, count(*) from odm2.samplingfeatureexternalidentifiers group by samplingfeatureexternalidentifier order by count(*) desc
```

| samplingfeatureexternalidentifier | count |
| :--- | :--- |
| 73526 | 3 |
| 73554 | 2 |
| 73564 | 2 |
| 73536 | 2 |
| 73562 | 2 |
| 73561 | 2 |
| 73556 | 2 |
| 73523 | 2 |
| 73555 | 2 |
| 73560 | 2 |
| 73570 | 2 |
 

example:

| bridgeid | samplingfeatureid | externalidentifiersystemid | samplingfeatureexternalidentifier | samplingfeatureexternalidentifieruri |
| :--- | :--- | :--- | :--- | :--- |
| 61400 | 61927 | 2 | 73526 | NULL |
| 61401 | 61928 | 2 | 73526 | NULL |
| 61402 | 61929 | 2 | 73526 | NULL |

[...]

PR #208, opened 2022-02-07
Updating water_sample and schemas according to schema changes due to project station sync

ZofiaRudjord:


api/src/odm2_postgres_api/queries/core_queries.py
```diff
@@ -77,9 +77,21 @@ async def find_unit(conn: asyncpg.connection, unit: UnitsCreate, raise_if_none=F
...
+async def find_site(conn: asyncpg.connection, projectstationcode: str, project_O_number: str) -> Optional[int]:
     result = await conn.fetchrow(
-        "select samplingfeatureid from samplingfeatures where samplingfeaturecode=$1 ",
+        "select * from samplingfeatures "
+        "join projectstations p on samplingfeatures.samplingfeatureid = p.samplingfeatureid "
+        "join projects pr on pr.projectid=p.projectid "
+        "where p.projectstationcode=$1 and $2 = any(pr.projectnumbers)",
+        projectstationcode,
+        project_O_number,
+    )
```
haakondr, line 88: this query can in theory return multiple rows in cases where multiple projects have the same onumber and these projects have the same projectstation codes. In such cases fetchrow will raise an `asyncpg.exceptions.CardinalityViolationError` error as it expects only one row returned. Similar to the issue we had in https://github.com/NIVANorge/odm2-postgres-api/issues/201.

This is probably quite unlikely but the schema allows for it. Just thought I'd mention it in case it happens. Could consider to explicitly catch this exception and print a user friendly error message

api/src/odm2_postgres_api/queries/core_queries.py
```diff
@@ -664,22 +676,12 @@ async def register_water_sample(conn: asyncpg.connection, data: schemas.WaterSam
...
-                schemas.Directive,
-                raise_if_none=True,
-            )
             ms_sample_data = schemas.ActionsCreate(
                 actiondescription="Registered water sample",
                 begindatetime=data.collection_datetime,
                 actiontypecv="Specimen collection",
                 methodcode="collect_sample",
                 isactionlead=True,
-                directiveids=[directive.directiveid],
```
haakondr, line 682: this means that actions are no longer associated with projects. I assume this was the previous way of modelling the project relationship and this is now expressed via the station -> project station. Is that correct?

haakondr (approved):

api/src/odm2_postgres_api/queries/core_queries.py
ZofiaRudjord: Yes, exactly, since now projects are linked the same as in nivabase via projectstaions to samplingfeature , directives are obsolete.

PR #210, opened 2022-02-08
Updating register_spectra such that new json is inserted if processing parameters changed.

ZofiaRudjord:
I have not tested the code locally yet. I will setup kafka sink now and do the testing.


ZofiaRudjord:
I tested it locally and it works as expected. Ok to merge @haakondr  ?

haakondr (approved):

